Issue #326, opened 2025-05-29
Create framework integration test suite for React, Vue, Solid, Svelte, etc.


schickling:
Create a comprehensive test suite to ensure all framework integrations (React, Vue, Solid, Svelte, etc.) have the same level of support for LiveStore. The test suite should cover:

- Proper lifecycle management: loading, unmounting, interrupting, error states, prop changes, etc.
- Using multiple LiveStore store instances in parallel within the same app
- Unit tests for individual framework-specific integrations (e.g. React hooks)

This will help ensure consistency and reliability across all supported frameworks.

PR #369, opened 2025-06-09
feat: Svelte adapter

### LINKED ISSUE #315: Svelte framework integration (similar to React/Solid)
It would be great to have a Svelte integration for LiveStore, similar to the existing React and Solid integrations. This would make it much easier for Svelte developers to use LiveStore in their projects and would help broaden the adoption of the library.

Additionally, Rich Harris (creator of Svelte) has mentioned that he might be interested in helping with the creation of a Svelte adapter, which could be a great opportunity to collaborate and ensure the integration is idiomatic for Svelte users.

Would love to discuss what an ideal Svelte integration would look like and how we can get started!

elliott-with-the-longest-name-on-github:
Closes #315 

Related to #326

This obviously still has rough edges (I think we need to wire up some OTEL stuff?), but it's a first working version hooked up to Svelte's reactivity system.

The API surface area for Svelte is actually quite small, as I'm not sure how much additional value we can bring:
- `createStore` is a _very_ thin wrapper around `createStorePromise`. It registers an effect to clean up the store when it's destroyed.
    - This has the side effect (ha ha) of forcing `createStore` to be called somewhere within an effect root, meaning it can always be cleaned up.
- We add a `createQuery` property to the store that takes a LiveStore query and returns a thunk wrapping the reactive version of that query's response. 
    - I tried a few different versions of this API, and this was the best one I could find. This allows you to use `$derived.by` to unwrap the value (+ allow reactive destructuring if that makes sense for your usecase), or just create the query and then call the result wherever you actually want to use it.

I'm not clear on whether there are additional value-adds we can create. I considered a standalone `createQuery` similar to what React has, but I don't really think it's worth it. Unwrapping promises in Svelte is pretty simple and easy (so store creation is easy), and sharing references to an object efficiently is also easy, so the convoluted context stuff React has to do really doesn't add any value. This also makes passing around multiple different store instances easier.

One thing I have thought of doing is exporting the `withSvelteApi` provider so that users can attach it to their LiveStore instances that might live outside of a Svelte context (and therefore cannot access effects for automatic cleanup). I'm not sure this is useful though, so it's probably something we could leave off and do later if it's clear there's a need.

I'd like feedback on the proposed API prior to investing more time in tests / OTEL / etc.

### Checklist
- [x] I grant to recipients of this Project distribution a perpetual,
non-exclusive, royalty-free, irrevocable copyright license to reproduce, prepare
derivative works of, publicly display, sublicense, and distribute this
Contribution and such derivative works.
- [x] I certify that I am legally entitled to grant this license, and that this
Contribution contains no content requiring a license from any third party.


.vscode/settings.json
```diff
@@ -28,7 +28,7 @@
   "files.exclude": {
     "**/prettier.config.cjs": true,
     "**/node_modules/**/dist": false,
-    "**/{examples,packages}/**/dist": true,
+    // "**/{examples,packages}/**/dist": true,
```
elliott-with-the-longest-name-on-github, line 31: TODO remove this -- it was preventing me from seeing dist so I could debug some setup issues

examples/src/web-todomvc-svelte/README.md
```diff
@@ -0,0 +1,38 @@
+# sv
```
elliott-with-the-longest-name-on-github, line 1: TODO easy to update for this example

themixednuts:
Thanks, just used this while I figured out how to make sync work with SvelteKit. This is an example repo https://github.com/themixednuts/sveltekit-livestore-cloudflare

timscodebase:
Is there an ETA for this?  Svelte is by far my favorite meta framework.

Sillyvan:
I was playing with this using async svelte so something like 
```ts
const store = await createStore({ adapter, schema, storeId: 'default' });
```

```html
<section class="todoapp">
	<Test {store} />
</section>
```

funny enough committing the events works fine. but deriving the state dosent work anymore. as in it never updates the state on the ui

schickling:
Thanks so much again for your fantastic work on this, Elliott, for kicking this off. [Rich has started a new pull request here](https://github.com/livestorejs/livestore/pull/849) which pushed us over the finish line. Thanks a lot again for all your efforts and help with this great contribution.

Issue #373, opened 2025-06-11
linearlite example: 404 from netlify on direct non-root url access


ldirer:
Directly accessing [https://web-linearlite.livestore.dev/board](https://web-linearlite.livestore.dev/board) returns a 404:

![Image](https://github.com/user-attachments/assets/bb0638b2-e639-46b6-b299-3d006b37bfc5)

This is only an issue when directly entering the full url: the root url loads the app just fine and navigation works.

I think it's possible that the `netlify.toml` file lives in a directory that is not picked up by the CLI (it's in `public/` and gets copied to `dist/`).

What I tried:

1. I ran the netlify deploy command with `--debug`
2. I added a symlink to the root folder of the example with `ln -s public/netlify.toml ./netlify.toml`
3. I ran the netlify deploy command with `--debug` again: it changed the output to include the config:
        
        ❯ Resolved config
        build:
          publish: /third_party/livestore/examples/src/web-linearlite
          publishOrigin: default
        redirects:
          - from: /*
            status: 200
            to: /index.html
        redirectsOrigin: config

This is a minor issue but it requires Netlify access to fix.

ldirer:
See https://github.com/livestorejs/livestore/pull/393#issuecomment-3048110421

PR #393, opened 2025-06-25
fix: misplaced netlify config for linearlite example causing occasional 404s

ldirer:
## Summary

This is a tiny fix for #373 .

## Related issues

- #373

## How to test

I just realized I could test this myself as the examples should be deployed by the CI for this PR.
I'll report on the fix.


ldirer:
Looks like the PR doesn't have access to the Netlify secret... which makes sense.
I was hoping to test the change in behavior on the snapshot-deployed linearlite example.

We could merge this into a in-repo branch and then have a PR to deploy the preview.

I'd say just trust in the change and merge it into main, the stakes are pretty low.

PR #398, opened 2025-06-27
`materialize(events, {})` -> `{} satisfies Materialize<typeof events>`

harrysolovay:
## Summary

Given that (A) events aren't used by `materialize` at runtime and (B) tables and events are defined as records, I think it makes sense for materializers to be defined via a `Materializers` type util and object literal. Ie.

```ts
const materializers: State.SQLite.Materializers<typeof events> = {
  todoCreated: ({ id, text, completed }) => {
    // ...
  },
  // ...
}
```

This also ensures users don't run into use before declaration errors.

```ts
const materializers = State.SQLite.materializers(events, {
  //                                             ~~~~~~
  //                                             ^ Block-scoped variable 'events' used before its declaration.ts(2448)
  // ...
})

const events = {
  // ...
}
```

Also added a small snippet to the materializer docs about materializer arm return types (other than the livestore builder).


harrysolovay:
A bit out of date at this point + not a necessary DX change in the first place

PR #502, opened 2025-07-22
Modified the Solid.js integration to work through useQuery and useStore hooks

joodaloop:
## Summary

This PR adds the functions `useQuery` and `useStore`, as well as the appropriate LiveStore context and provider code to update the Solid integration to match the existing React functionality. 

It switches to using Solid stores (with reconcile) instead of a coarse signal of the previous implementation. It also allows for using multiple stores instead of the singleton pattern enforced by the previous `getStore` function.

The Effect code to manage the store lifecycle within the provider is written with help from Gemini 2.5 Pro. The query caching in `useQuery` was adapted from the React version of useQuery.

Note: I will update the examples and docs in a future PR if these new methods prove acceptable.

### DIFF
--- a/packages/@livestore/solid/src/LiveStoreContext.ts
+++ b/packages/@livestore/solid/src/LiveStoreContext.ts
@@ -1,0 +1,16 @@
+import { createContext, useContext } from "solid-js";
+import type { Store, LiveStoreSchema } from "@livestore/livestore";
+
+export type LiveStoreContextValue = {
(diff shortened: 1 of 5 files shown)


schickling:
Thanks a lot for this PR! The argument makes a lot of sense to me.

Would love to see some tests similar to the React integration package as there are many nuanced life-cycle scenarios and nuanced cases that should be tested.

joodaloop:
You're right, will see what I can do there!

PR #532, opened 2025-07-31
Fix the livestore.worker.ts path in guides

kuldar:
Currently in the "Getting Started" guide for "React Web", the instructions tell you to create the `livestore.worker.ts` file into `src/livestore` directory, but the example files referenced in next steps have the `livestore.worker.ts` file in `src` directory instead.

Not sure which is the preferred location, but since it was easier to fix the "Getting Started" page instructions than the actual source code for the examples, I kept the `src` directory approach.

Screenshot 2025-07-31 at 21 02 39
Screenshot 2025-07-31 at 21 02 20

### DIFF
--- a/docs/src/content/docs/getting-started/react-web.mdx
+++ b/docs/src/content/docs/getting-started/react-web.mdx
@@ -140,9 +140,9 @@ Here's an example schema:
 
 ## Create the LiveStore Worker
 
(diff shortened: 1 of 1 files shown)


schickling:
Sorry for the delay on this. This is now fixed in the latest version of the docs.

PR #576, opened 2025-08-25
demo on issue commiting events inside subscribe() 

debuggingfuture:


schickling:
Bug confirmed. Thanks for the repro.

PR #585, opened 2025-08-28
docs(rfc): Multi-Store API Design

schickling:
This introduces [**RFC 0001: Multi-Store API Design**](https://github.com/livestorejs/livestore/blob/investigating-multi-store-usage-patterns/contributor-docs/rfcs/0001-multi-store-api-design.md). A proposal for supporting multiple store instances within a single application.

## Key Areas for Reviewers
If you're short on time, review these key items:

1. **API ergonomics**: e.g. `StoreRegistry`, `storeOptions({...})` and `useStore({...})` usage
2. [**Single store vs. multi-store API**](https://github.com/livestorejs/livestore/blob/investigating-multi-store-usage-patterns/contributor-docs/rfcs/0001-multi-store-api-design.md#1-how-should-we-reconcile-the-single-store-api-with-multi-store): How should the single-store usage look like with the introduction of the multi-store API
3. [**Configuration cascade**](https://github.com/livestorejs/livestore/blob/investigating-multi-store-usage-patterns/contributor-docs/rfcs/0001-multi-store-api-design.md#2-which-store-options-should-be-available-at-each-layer-and-how-should-values-merge): Which options should be available in each layer (StoreRegistry → call-site) and how should they override?
4. **Framework support**: Any concerns about adapting this beyond React (Vue, Svelte, etc.)?

## Example App

See #774 for practical demonstrations of five usage patterns: single store, chained stores, independent stores, multi-instance, and recursive composition.

## Try it now
An experimental version is already available in `0.4.0-dev.14` under `@livestore/react/experimental`!

PR: https://github.com/livestorejs/livestore/pull/764

### RFC Process

To support this and future proposals, this PR also establishes the RFC process:

- **RFC template** (`docs/src/content/rfcs/0000-template.md`)
- **RFC index** (`docs/src/content/rfcs/index.md`) with process documentation
- **Updated contributing guide** to reference RFCs for significant changes


joodaloop:
Big fan of this API, especially how well it can play with Suspense!

My only comment is a semantic one I've mentioned earlier, suggesting calling the hooks `useLiveQuery` and `useLiveStore` instead of the more generic `useQuery` (used by the very popular Tanstack Query) and `useStore` (Svelte and Solid both have primitives called "store").

This would allow other framework integrations to follow the same names as the React one, while minimizing name conflict. It might also make it less likely to confuse LLMs who are very used to the original `useQuery`.

slashv:
@IGassmann Thanks for your response. Happy to collaborate on this going forward.

> * Made the `<XStoreProvider>` provider responsible for suspending its component tree instead of the `useXStore()` hook.

Conceptually this makes sense but I have to evaluate how this would work in Vue. The XStoreProvider would need to become an async component and ensure the store is loaded before the inner useXStore can be called synchronously. I believe I tested this approach at one stage and ran into an issue but I will give it another shot. It's possible frameworks will need to diverge a bit here but let's see.

> * Updated implementation and examples to default `storeId` to the store's `name` when not specified.

Ok, I can add the same behaviour for Vue. The potential downside is that it conceals the concept of the storeId which I found rather useful to understand when learning how LiveStore works but I like that it can make the most common scenario of a top level store cleaner which I think is a good thing. It's a trade-off between allowing simplicity and enforcing clarity.

> * Only allow to pass a `storeId` at the Provider level and not when calling `createStoreContext`.
>   * Allowing `createStoreContext` to bake in a `storeld` buys us a default instance identifier, but it also introduces a hidden coupling between the factory and every Provider we render later. In practice the actual store ID nearly always depends on runtime state (project/workspace IDs, etc.), so pushing callers to pass it through the Provider keeps that wiring explicit.

I agree. From the examples I worked through this is the most common approach and making it explicit clarifies the responsibility of the provider as the creator of the store. The exception would be a single top level store where it might be cleaner passing in storeId to `createStoreContext` but even then I would agree with your approach.

If the idea is to make `createStoreContext` the default and deprecate `LiveStoreProvider` I'm partial towards agreeing with @drowrin that `defineStore` would be a better name. Considering we are both using the pattern he suggests with `export const [WorkspaceStoreProvider, useWorkspaceStore] = defineStore({...` (and not using `React.use`) the concept of a context is perhaps less relevant. I also found it more natural to name the file where the store is defined `store.ts` which would align with `defineStore`.

The reason I asked about common usage scenarios for multiple stores is that the workspace/project/issue example could be done with a single store so it would be good to give reasons why splitting into separate stores is beneficial. Is it simply to facilitate code segmentation or are there other obvious benefits for topics like permissions handling or performance?

IGassmann:
The current proposal, by design, bounds every store instance's lifetime to its enclosing provider (e.g. `<WorkspaceStoreProvider>`) boundary. The only way to load/unload a store is to mount/unmount the provider. That coupling makes partial synchronization, where we need to to spin stores up and down on demand, effectively unattainable, and it blocks any centralized policy for deduplication, invalidation, or preloading of stores:

- **Preloading before mount**: the only store registration path runs inside `StoreRegistrar`’s `useEffect`, which executes only after the provider commits. There’s no API to create a store ahead of time (e.g., in a router loader) and then let components consume it when they mount. Without preloading, initial reads always suspend until the mount + effect completes, and we can’t hydrate stores opportunistically before navigating. Considering the non‑negligible time it takes to boot a store, this greatly affect UX.
- **Deduplication gaps**: mounting two providers with the same `storeId` simply creates two independent LiveStoreProvider instances. Because the per-provider registry never escapes that provider, we can’t detect the duplicate nor reuse the first store instance. The result is double loading, divergent state, and no ability to reconcile which copy should be kept when memory is tight.
-  **Can't warm cache stores**: Store instances unload as soon as its provider unmounts. There’s no way to keep a store “warm” in cache after its UI goes away—say, when we navigate off a route but want a fast return—because lifetime is hard-wired to the provider’s presence in the tree. That blocks classic prefetch-and-cache patterns: every time the component remounts we pay the full reinitialization cost instead of reusing a cached store instance.
- **Cross-type ops & invalidation**: `createStoreContext` spins up a brand-new `RegistryContext` for every store type. That isolates each store type in its own registry, so there’s no single place to invalidate all stores after a shared event (e.g., logout, workspace purge) or to orchestrate cross-type cascades. A partial-sync cache needs one coordinator to evict or refresh related stores together; the current design forces we to walk every mounted provider manually instead.

To solve these issues, I believe we need to replace the per-provider registry with a single global registry/cache that the whole app can talk to. I've put together a new API proposal that does just that (with three variants):

- [Global Registry + React Bindings ](https://github.com/livestorejs/livestore/blob/97fa86a485ef99ea610f7f5751882da7ca3cf96a/wip/multi-store/api-design-global-registry-a.md)
- [Global Registry + Universal Hook](https://github.com/livestorejs/livestore/blob/97fa86a485ef99ea610f7f5751882da7ca3cf96a/wip/multi-store/api-design-global-registry-b.md)
- [Global Registry + Promise-first](https://github.com/livestorejs/livestore/blob/97fa86a485ef99ea610f7f5751882da7ca3cf96a/wip/multi-store/api-design-global-registry-c.md)

This fixes the aforementioned issues, but brings its own challenges. It pretty much requires implementing a caching layer similar to TanStack Query with eviction policies, preloading, invalidation, etc

PR #597, opened 2025-09-02
fix: use proper git reference in build-example-tiged workflow

schickling:
## Summary
- Fix CI failure in `build-example-tiged` workflow caused by invalid git reference
- Replace `pnpm dlx` with `bunx` for consistency

## Problem
The workflow was using `github.ref_name` which evaluates to `"123/merge"` in PR contexts - an invalid git reference that `tiged` cannot fetch from the repository. This caused the workflow to consistently fail on pull requests.

**Example failing CI run:** https://github.com/livestorejs/livestore/actions/runs/17386194093/job/49353595476#step:5:2

The specific error occurred in the "Copy example app" step where `tiged` could not fetch the repository using the invalid reference.

## Solution
- Use `github.event.pull_request.head.sha || github.sha` to provide the actual commit SHA that `tiged` can successfully fetch
- Replace `pnpm dlx` with `bunx` for consistency with project tooling preferences
- Update comment to accurately explain the approach

## Test plan
- [ ] Verify workflow runs successfully on this PR
- [ ] Confirm `build-example-tiged` job completes without errors
- [ ] Test that both PR and push contexts work correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,10 +246,10 @@ jobs:
             jq -r '.[0].dependencies | keys | join(" ")' \
           )" >> $GITHUB_ENV
       - name: Copy example app
-        # We use `github.ref_name` instead of `github.sha` because, when a workflow is triggered by a pull request,
-        # `github.sha` refers to a temporary commit SHA that can become inaccessible in some contexts.
+        # We use the PR head SHA (or push SHA) to ensure tiged can fetch the exact commit.
+        # In PR context, github.ref_name would be "123/merge" which isn't a valid git reference.
         # See https://www.kenmuse.com/blog/the-many-shas-of-a-github-pull-request/
-        run: pnpm dlx tiged github:${{ github.repository }}/${{ env.APP_PATH }}#${{ github.ref_name }} ${{ runner.temp }}/${{ env.APP_PATH }}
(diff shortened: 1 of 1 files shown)


IGassmann:
> Thanks a lot for providing this analysis. This makes sense overall. I'll close the issue for now. That being said, I'm still not 100% sure why it wouldn't be safe to point to the git commit hashes.

To clarify, the issue isn't in using commit hashes. `github.event.pull_request.head.ref` and `github.event.pull_request.head.sha` are equivalent (branch name vs. last commit hash of the same branch), just like `github.ref` and `github.sha` are equivalent for the merge branch.

The key distinction is which commit hash we'd be pointing to. In an open pull request, there are three branches:

  1. **The merge branch** (referenced by `github.ref` / `github.sha`) - GitHub's temporary merge of head → base
  2. **The source (HEAD) branch** (referenced by `github.event.pull_request.head.ref` /` github.event.pull_request.head.sha`) - the PR feature branch (e.g., `fix-node-sync-test`)
  3. **The base branch** (referenced by  `github.event.pull_request.base.ref` / `github.event.pull_request.base.sha`) - the target branch (e.g., `dev`)

  The problem with switching to `github.event.pull_request.head.sha` (or `github.event.pull_request.head.ref`) is that it would make `tiged` point to the source branch's commit instead of the merge branch's commit. This creates a mismatch because:

  - All our jobs build and test against the merge branch (what the code looks like after merging), not the source branch
  - The snapshot versions are built from the code in the merge branch, not the source branch

PR #622, opened 2025-09-03
[H001] Default resume-on-advance (guarded by parked state)

schickling:
# Summary
Make resume-on-advance the default fix for H001: when pushing is parked due to ServerAhead, restart pushing on upstream advance (not only on rebase). Guarded by a parked-state flag to avoid unnecessary restarts.

# Details
- Track `pushParked` in LeaderSyncProcessor.
- `backgroundBackendPushing` accepts `onPark()` and sets it when handling ServerAhead by parking (Effect.never).
- On pull `rebase`, restart pushing as before.
- On pull `advance`, restart pushing if and only if `pushParked` is true.

# Rationale
CI frequently produces advance-only sequences; previously the push loop stayed parked until a rebase occurred, causing stalls/timeouts. This change safely resumes pushing on advance when needed.

# Safety
- Default on. Can be disabled via env in a follow-up PR if needed.
- Minimal surface area; no protocol semantics changed.
- Keeps current rebase path behavior.

# Follow-up
- Remove extra diagnostics added in hypothesis branches once this lands.
- Consider adding a focused regression test for advance-only resumption.

### DIFF
--- a/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
+++ b/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
@@ -831,7 +831,8 @@ const backgroundBackendPushing = ({
             const tag = String((pushResult.left.cause as any)._tag ?? '')
             if (tag.includes('UnexpectedError')) {
               const raw = String((pushResult.left.cause as any).cause ?? pushResult.left.cause)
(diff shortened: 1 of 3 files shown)


schickling:
Closing debug PR per request; pausing clean fix for now.

PR #652, opened 2025-09-17
Fix the types of `Vitest.makeWithTestCtx`

### LINKED ISSUE #633: TypeScript fails to catch Layer type mismatches with generic defaults (tsc passes but VS Code errors)
## Problem Description

TypeScript's command-line compiler (`tsc`) fails to catch certain type errors in Layer compositions when using functions with default generic parameters, while VS Code's TypeScript language server correctly identifies these errors. This creates a situation where CI/build processes may pass while developers see errors in their IDE.

## Root Cause

The issue occurs when:
1. A function has generic parameters with default values (e.g., `<R1 = never, E1 = never>`)
2. TypeScript's type inference becomes too permissive with these defaults
3. The error only surfaces with `skipLibCheck: false`, which is not feasible for most projects

## Reproduction Steps

1. **Create a test file that uses `WranglerDevServerService` with `makeWithTestCtx`:**

```typescript
// packages/@livestore/common-cf/src/do-rpc/rpc.test.ts
import { Layer } from '@livestore/utils/effect'
import { PlatformNode } from '@livestore/utils/node'
import { WranglerDevServerService } from '@livestore/utils-dev/node'
import { Vitest } from '@livestore/utils-dev/node-vitest'

const withWranglerTest = Vitest.makeWithTestCtx({
  timeout: 60_000,
  makeLayer: () =>
    WranglerDevServerService.Default({
      cwd: `${import.meta.dirname}/test-fixtures`,
    }).pipe(Layer.provide(PlatformNode.NodeContext.layer)),
    // Missing: Layer.provide(FetchHttpClient.layer)
})
```

2. **Run TypeScript compilation:**
```bash
direnv exec . mono ts
# OR
direnv exec . tsc --noEmit
```
Result: ✅ **No errors reported**

3. **Open the file in VS Code:**
Result: ❌ **VS Code shows error on line with `makeLayer:`**
```
Type '() => Layer.Layer<WranglerDevServerService, WranglerDevServerError, HttpClient>' is not assignable to type '(testContext: TestContext) => Layer<WranglerDevServerService, WranglerDevServerError, Scope>'.
  Type 'Layer<WranglerDevServerService, WranglerDevServerError, HttpClient>' is not assignable to type 'Layer<WranglerDevServerService, WranglerDevServerError, Scope>'.
```

4. **To make `tsc` catch the error, you need to either:**
   - Set `skipLibCheck: false` in tsconfig.json (not feasible with 1000+ .d.ts files)
   - Provide explicit type parameters: `Vitest.makeWithTestCtx<WranglerDevServerService, never>({...})`
   - Add explicit return type to makeLayer function

## Impact

- This issue can cause runtime failures that aren't caught during build/CI
- Developers relying on VS Code see errors that CI doesn't catch
- Affects any code using `makeWithTestCtx` with Layer compositions

## Environment

- TypeScript: 5.9.2
- Effect: 3.17.9
- skipLibCheck: true (standard configuration)

## Additional Context

The issue is specifically about TypeScript's generic inference with default parameters being too permissive. This is a known TypeScript behavior where generic inference with defaults can be less strict than explicit type parameters.

Related files affected:
- `packages/@livestore/common-cf/src/do-rpc/rpc.test.ts`
- `packages/@livestore/common-cf/src/ws-rpc/ws-rpc.test.ts`
- `packages/@livestore/utils-dev/src/node-vitest/Vitest.ts`

IMax153:
## Summary

Explain the **what** and **why** of the changes.

## Related issues

- Closes #633

### DIFF
--- a/packages/@livestore/utils-dev/src/node-vitest/Vitest.ts
+++ b/packages/@livestore/utils-dev/src/node-vitest/Vitest.ts
@@ -20,7 +20,7 @@ export * from '@effect/vitest'
 
 export const DEBUGGER_ACTIVE = Boolean(process.env.DEBUGGER_ACTIVE ?? inspector.url() !== undefined)
 
-export const makeWithTestCtx: <ROut, E1, RIn>(
+export const makeWithTestCtx: <ROut = never, E1 = never, RIn = never>(
   ctxParams: WithTestCtxParams<ROut, E1, RIn>,
 ) => (testContext: Vitest.TestContext) => <A, E, R>(
   self: Effect.Effect<A, E, R>,


schickling:
Thank you!

PR #713, opened 2025-10-01
Fail fast when DO WebSocket drops mid-pagination

schickling:
## Context
We’re mid-way through H001 (Cloudflare DO WebSocket large-history stall). The long-term goal is to resume pagination automatically when the DO closes the socket mid-stream. This PR is an interim milestone so CI stops timing out while we design the resume path.

## Summary
- bubble `PullStreamInterruptedError` when the DO WebSocket closes with remaining history so the suite fails fast instead of hanging
- record chunk/event totals and log RPC scope exits to correlate DO-side interruptions while the investigation continues
- update the Cloudflare DO sync-provider spec to assert this temporary fail-fast behaviour

## Next steps
- capture the new `TMP ws scope close …` and client close logs on fresh fail-fast runs to pinpoint who initiates the clean close
- prototype resume logic that reissues `pull` with the last cursor once the worker reconnects
- strip the temporary instrumentation once resume semantics are in place

## Testing
- `CI=1 direnv exec . vitest run tests/sync-provider/src/sync-provider.test.ts --testNamePattern "'Cloudflare WebSocket (DO)'"`

### DIFF
--- a/packages/@livestore/common-cf/src/ws-rpc/ws-rpc-server.ts
+++ b/packages/@livestore/common-cf/src/ws-rpc/ws-rpc-server.ts
@@ -17,6 +17,7 @@
 
 import { notYetImplemented, omitUndefineds } from '@livestore/utils'
 import {
+  Cause,
   constVoid,
   Effect,
   Exit,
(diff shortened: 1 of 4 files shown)


schickling:
Closing for now as we're tracking this in an open issue.

Issue #513, opened 2025-07-25
Implement comprehensive LiveStore events API with streaming and advanced filtering


schickling:
## Summary

Add a complete events API to LiveStore that allows users to access, filter, and stream historical events with configurable sync level visibility and advanced filtering options.

## Motivation

Currently, LiveStore lacks a comprehensive way to access historical events from stores. The store APIs (`store.events()` and `store.eventsStream()`) are already exposed but currently throw "not implemented" errors. Users need the ability to:
- Stream events efficiently from the eventlog database
- Filter events by various criteria (event names, client IDs, session IDs, time ranges)
- Control visibility based on sync levels (client, leader, backend)
- Handle large event histories with proper batching and performance

## Proposed API

```typescript
// Async iterable interface
for await (const event of store.events(options)) {
  // Process events
}

// Effect Stream interface  
const eventsStream = store.eventsStream(options)
```

## Implementation Requirements

- [ ] Core events API (`store.events()` and `store.eventsStream()`)
- [ ] Efficient streaming from SQLite eventlog with batching
- [ ] Advanced filtering with SQL-level optimization
- [ ] Leader thread proxy streaming interface
- [ ] Adapter implementations across all platforms
- [ ] Comprehensive test coverage

## Related Issues

- #465 Event Snapshotting
- #353 Improve handling of unknown livestore events

## Implementation

See PR #514

IGassmann:
@slashv @schickling Is this completed now that https://github.com/livestorejs/livestore/pull/753 has been merged?

PR #753, opened 2025-10-16
Fix 513 event streaming v2

slashv:
Supersedes #513
includes and extends it's commits

Stream events from the eventlog via `store.events` as an AsyncIterable or `store.eventStream` as an Effect.Stream. Currently only supports streaming events confirmed by the sync backend.

Allows specifying range of events to stream:
- `since?`: `EventSequenceNumber.Client.Composite` (exclusive)
- `until?`: `EventSequenceNumber.Client.Composite` (inclusive)

If an until marker is supplied the stream finalizes upon reaching it. Defaults to tracking upstreamHead via `syncState`

Allows filtering events by:
 - `filter`: event types
 - `clientIds`: client identifiers
 - `sessionIds`: session identifiers

`packages/@livestore/common/src/leader-thread/stream-events.ts`
Core streaming logic is handled in `streamEventsWithSyncState` which gets imported and used in adapters to allow stream processing to run on the leader thread.

`packages/@livestore/common/src/leader-thread/eventlog.ts`
The eventlog queries get handled in `getEventsFromEventlog`

## Usage examples
```ts
// Run once
for await (const event of store.events(filter: ['todoCompleted'])) {
  console.log('event from leader', event)
}

// Continuos stream
const iterator = store.events()[Symbol.asyncIterator]()
try {
  while (true) {
    const { value, done } = await iterator.next()
    if (done) break
    console.log('event from stream:', value)
  }
} finally {
  await iterator.return?.()
}
```

## Implementation notes

The **batchSize** parameter defines the maximum size of query batches for each query. The default value has been set to 100 as this was measured to provide the best performance. We might want to look closer at tuning this parameter when we have memory profiling in the performance test working but for now it's user configurable and might be platform and use-case specific for optimal value. Only minor performance difference was measured between 10 and 1000.

A few other implementation approaches were considered. These might still be of value at a later stage to further improve performance but before session, and leader level streams with support for unconfirmed events is completed we decided that the added complexity is premature.

Fetch plans into a Sink
https://gist.github.com/slashv/f1223689f2d1171d2eeb60a2823f4c7c

Fetch plans into sink and decompose into windows
https://gist.github.com/slashv/a8f55f50121c080937f42e44b4039ac8

Mailbox and Latch approach (suggestion by Tim Smart)
https://gist.github.com/slashv/d6b12395c85415bf0d3363372a1636c3

## Future improvements

- Support streaming unconfirmed events
  - Leader level
  - Session level
- Support streaming clientOnly events
- Improve performance test
  - Automatically generating snapshots for testing higher event-counts
  - CPU usage reporting
  - Memory usage reporting (requires additional tooling to be useful on worker thread)

### Unit tests

`tests/package-common/src/leader-thread/stream-events.test.ts`

- [x] Stream progresses as upstream head advances
- [x] Events get filtered by name
- [x] Events streamed until specific upstream head includes it then finalizes
- [x] Events streamed since local head doesn't include local head
- [x] Events get filtered by client ID
- [x] Events get filtered by session ID
- [x] Client only events get filtered out
- [x] Correctly streams events across batches (property based test)

### Integration tests

`packages/@livestore/livestore/src/store/store-eventstream.test.ts`

- [x] Stream resumes when reconnected to sync backend

### Performance tests

`tests/perf-eventlog/tests/suites/event-streaming.test.ts`

- [x] Latency test (1000 events)
- [x] Large event log 10K (requires pre-generating snapshot)
- [x] Large event log 100K (requires pre-generating snapshot)

## Other changes

In order to allow loading both state and eventlog db snapshots `LSD.Leader.LoadDatabaseFile.Request` in `packages/@livestore/common/src/leader-thread/leader-worker-devtools.ts` was extended to support passing in both databases at once. This was required for the performance testing large event batches.


slashv:
Couple of comments / questions from my initial review:

- Should the eventStream when snapshot=true emit.end() after all events have been iterated over? When it didn't the test would have to take(n) in the caller instead of runDrain of the stream but in a real usage scenario that number isn't known and I would expect snapshot to be finite and have an end.

- Existing tests used createMockSyncProcessor but that doesn't test the store logic. I created a new initial test that uses makeTodoMvc which I saw used in other places. Let me know if you think this is the right approach.

slashv:
After discussion with @schickling I've started focusing on defining a test plan for different scenarios that might occur. A starting point for this can be found in [events-test-plan.md](https://github.com/slashv/livestore/blob/fix-513-event-streaming/docs/events-test-plan.md).

In order to meaningfully contribute to this feature I had to do some research into understanding how the event system works. I've documented my understanding of it in [this document](https://github.com/slashv/livestore/blob/fix-513-event-streaming/docs/understanding-events.md). This is just temporary doc but if any part of this is wrong or I'm missing some crucial element that would be very helpful to know.

In the test plan I also included some reflections which I'll paste here for easier reference. These primarily deal with the api design but are relevant to deciding what testing scenarios to account for.

**Only streaming confirmed events**

Streaming only confirmed events as we discussed removes the current value of minSyncLevel client (which merges pending events) and leaves backend as the only level that guarantees confirmation which is not useful in offline scenarios. At least if my understanding of what a confirmed event counts as is correct.

**Dynamically pushing `until` marker to align with upstream head**

I would think users expect a backend-level stream to follow the advancing head without manual restarts. Today `until` is fixed at stream start at backend level (and client but less relevant since merges pending sync state which is related to point above)

**Starting stream from local head**

Starting from the local head already works via a `since` cursor, but it would probably be beneficial to document how this can be done since I can imagine it being a common scenario where users wants to take actions only on new events and not the entire event log.

**Potential issue of missed events due to rebase mid-stream when batching**

Batch 1 might read `[e41, e42]` with `since = e39` and `offset = 0`. If a rebase then replaces those rows with `[e41′, e42′]`, the next query runs with the old `offset = 2`, so it jumps straight to `[e43, …]` and never surfaces the replacements. This mirrors `syncstate.test.ts` (“should rebase single client event to end”), where upstream injects `[e1_1, e1_2, e1_3, e2_0]` ahead of a pending event; the eventlog grows, but the streaming query still only sees the tail unless the consumer rewinds to a cursor ≤ `e1_1`.

schickling (approved):

schickling:
Incredible work! So excited to see this land!

PR #783, opened 2025-10-23
feat: re-write solid bindings to match react's API

bigmistqke:
## Problem

The API of react and solid were growing apart from each other, with solid's API not receiving any recent updates.

After a discussion with @schickling we came to these guidelines for the solid bindings:
1. focus on minimizing maintenance, each LOC counts
2. create as much possibility for code sharing in between react and solid's API
3. no `undefined` checks when accessing the store, block instead

## Solution

To address 1. I re-wrote the solid bindings to be a direct fork of react's package.

In the current state of the PR I ported everything, without too much consideration of its relevance in the context of solid e.g. I ported `RcResource` too, which is maybe only really necessary for strict mode react? If we can remove `RcResource`, we can also remove `useQueryRef` too.

A later commit will look if we can move some code to the `@livestore/shared`package for 2. but I already abstracted some code out of the components (the effect-code setting up the livestore-store).

For 3. I currently opted to copy the behavior of the react package, although solidjs' general philosophy is to prefer non-blocking async, to prevent waterfalls where possible.

In solid 1.x this means a lot of `undefined` checks, but in solid 2.x this DX annoyance is addressed by making async signals throw when they are not yet resolved, which would open up the door again to remove the initial block.

## Validation

I vibeported the tests with claude. They are very close to react's tests.

I currently rely on snapshots I generated from the solid tests, which is a bit dangerous because it relies on the snapshots being correct and I have too little knowledge about the livestore internals to verify the snapshots.


bigmistqke:
continued https://github.com/livestorejs/livestore/pull/958

PR #849, opened 2025-11-22
@livestore/svelte

Rich-Harris:
Similar to #369 (cc @elliott-with-the-longest-name-on-github), this adds a `@livestore/svelte` package that provides a way for Svelte developers to use LiveStore in their apps. It takes a slightly different approach — rather than adding a new `store.createQuery` method, it essentially just monkey-patches `store.query` to make it reactive in a Svelte context. As such, there is no additional API surface area, and you can use it inline, like so:

```svelte
{#each store.query(visibleTodos$) as todo (todo.id)}
  <li>...</li>
{/each}
```

If `store.query(...)` is called in a non-reactive context (e.g. an event handler, or while initializing a component) it's as though you just called the underlying method. But if you call it in a reactive context (a `$derived`, or an expression in the template) then it will call `store.subscribe(...)` and cause the expression to re-run when the query value changes.

Another difference with #369 is that `createStore` can be called outside an effect. I'm in two minds about this — it's very convenient if you can do `export const store = await createStore(...)` from a module, but it does mean that the user is responsible for memory management if a store is created outside the component tree. It's possible that a more explicit distinction between those two modes of use would be preferable.

## TODO

- [ ] figure out the type errors
- [ ] add some tests
- [ ] decide whether or not we like this API

## Alternative API

One wrinkle is that `store.query(...)` can take a `QueryBuilder`, but `store.subscribe(...)` doesn't do anything with it. Which means that if you have something like this...

```svelte
<span class="todo-count">
  {store.query(tables.todos.count().where({ completed: false, deletedAt: null })} items left
</span>
```

...it will display the correct number (since it's just calling the underlying `query` method) but when it changes, nothing will happen. I don't know if that's an indication that `store.subscribe` _should_ be able to subscribe to a `QueryBuilder` (I don't totally understand the difference between it and `LiveQuery`/`LiveQueryDef`/`SignalDef` etc), or if this isn't actually the right API for Svelte, and we should instead do something more like #369, or perhaps like this:

```ts
const incomplete = new LiveQuery(
  store,
  tables.todos.count().where({ completed: false, deletedAt: null }),
  { label: 'incompleteCount' }
);
```

```svelte
<span class="todo-count">{incomplete.current} items left</span>
```

This is arguably more Svelte-idiomatic, and is more efficient insofar as we don't need to keep recreating the subscription on every change. (The reactivity trick in this PR is a little bit outré, if perfectly valid.) It is slightly more verbose for the developer though, and involves a new API.


arnolicious:
Just out of curiousity, this would be a client-only affair, correct?
In a non-spa sveltekit context, I assume it would be difficult to have livestore run both on the server and the client, or have it run on the server and somehow pass the reactive queries and events back and forth between client and server?

schickling:
Yes, for now the LiveStore/Svelte integration is only working on the client side. However, bringing SSR and server-ish support for LiveStore is certainly a topic we hope to explore soon as well.

Issue #844, opened 2025-11-22
[Bug]: ElectricSQL example doesn't work


markmiro:
### Bug Description

I can't get the electric SQL example to run.

https://github.com/livestorejs/livestore/tree/dev/examples/web-todomvc-sync-electric

### Minimal Reproduction

Minimal reproduction: follow docs in the `examples/web-todomvc-sync-electric` README.

### LiveStore Version

0.4.0-dev.17

### Environment

```markdown
- OS: macOS (darwin 24.6.0) on Apple Silicon
- Node.js version: v24.2.0
- Package manager: bun (1.3.3)
- Runtime: Browser (primary) / Node.js
```

### Expected Behavior

README instructions to work

### Actual Behavior

Various issues that I can get around to varying levels of success. `docker compose up` runs. Packages install. The app displays in the browser and I can interact with it. `bun dev`, however, shows errors indicating the ElectricSQL isn't working when running with `pnpm`.

### Error Logs/Screenshots

When attempting with bun:
```shell
❯ bun dev
$ vite dev
```
(it hangs)

When attempting with pnpm:
```shell
❯ pnpm dev

> livestore-example-web-todomvc-sync-electric@0.0.0 dev /Users/mm/p/livestore/examples/web-todomvc-sync-electric
> vite dev


  VITE v7.1.9  ready in 671 ms

  ➜  Local:   http://localhost:60001/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
[vite] connected.
[baseline-browser-mapping] The data in this module is over two months old.  To ensure accurate Baseline data, please update: `npm i baseline-browser-mapping@latest -D`
PostgresError: role "postgres" does not exist
    at ErrorResponse (file:///Users/mm/p/livestore/node_modules/.pnpm/postgres@3.4.7/node_modules/postgres/src/connection.js:794:26)
    ... 10 lines matching cause stack trace ...
    at Object.migrate (/Users/mm/p/livestore/examples/web-todomvc-sync-electric/src/server/db.ts:20:5) {
  cause: PostgresError: role "postgres" does not exist
      at ErrorResponse (file:///Users/mm/p/livestore/node_modules/.pnpm/postgres@3.4.7/node_modules/postgres/src/connection.js:794:26)
      at handle (file:///Users/mm/p/livestore/node_modules/.pnpm/postgres@3.4.7/node_modules/postgres/src/connection.js:480:6)
[...]

PR #853, opened 2025-11-25
Fix ElectricSQL Example for bun

markmiro:
## Problem

ElectricSQL example doesn't work if you follow README.md instructions.

## Solution

Looks like the example broke after switching from `pnpm` to `bun`.

Adding `fsevents` addresses:

```shell
❯ bun dev
$ vite dev

  VITE v7.2.4  ready in 1265 ms

  ➜  Local:   http://localhost:60001/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
✘ [ERROR] No loader is configured for ".node" files: node_modules/fsevents/fsevents.node

    node_modules/fsevents/fsevents.js:13:23:
      13 │ const Native = require("./fsevents.node");
         ╵                        ~~~~~~~~~~~~~~~~~

/Users/mm/p/livestore/examples/web-todomvc-sync-electric/node_modules/esbuild/lib/main.js:1467
  let error = new Error(text);
              ^

Error: Error during dependency optimization:

✘ [ERROR] No loader is configured for ".node" files: node_modules/fsevents/fsevents.node

    node_modules/fsevents/fsevents.js:13:23:
      13 │ const Native = require("./fsevents.node");
         ╵                        ~~~~~~~~~~~~~~~~~


    at failureErrorWithLog (/Users/mm/p/livestore/examples/web-todomvc-sync-electric/node_modules/esbuild/lib/main.js:1467:15)
    at /Users/mm/p/livestore/examples/web-todomvc-sync-electric/node_modules/esbuild/lib/main.js:926:25
    at /Users/mm/p/livestore/examples/web-todomvc-sync-electric/node_modules/esbuild/lib/main.js:1345:9
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}

Node.js v24.2.0
error: script "dev" exited with code 1
```

The `package.json` change addresses:

```shell
❯ bun dev
$ vite dev

  VITE v7.2.4  ready in 763 ms

  ➜  Local:   http://localhost:60001/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
[vite] connected.
Error in renderToPipeableStream: TypeError: router.serverSsr.takeBufferedScripts is not a function
    at Scripts (/Users/mm/p/livestore/examples/web-todomvc-sync-electric/node_modules/@tanstack/react-router/src/Scripts.tsx:68:45)
```

After the above, I was getting the error `PostgresError: role "postgres" does not exist` when running `bun dev`. Turns out I had Postgres running locally and I didn't realize it. It was not an issue with my local setup rather than the repo. It could be useful for the example to give a better example if it's connecting to the wrong postgres DB.

`bun test` was failing, and fixed by running `bun run test`. I added documentation for that. Updated playwright config to use `bun` even though I believe it works without that change.

## Validation

1. Deleted `node_modules`
1. Ran `bun i`
1. Ran `bun dev`
1. Opened browser to validate that the page loads
1. Added and removed items. Opened incognito window with the same livestore ID, added and removed more items, confirmed syncing between the two

## Related issues

- #844


markmiro:
`bun test` fails with:
```
❯ bun test
bun test v1.3.3 (274e01c7)

tests/todomvc.spec.ts:

# Unhandled error between tests
-------------------------------
220 |       suite._timeout = timeout;
221 |       return;
222 |     }
223 |     const testInfo = (0, import_globals.currentTestInfo)();
224 |     if (!testInfo)
225 |       throw new Error(`test.setTimeout() can only be called from a test`);
                      ^
error: test.setTimeout() can only be called from a test
      at _setTimeout (/Users/mm/p/livestore/examples/web-todomvc-sync-electric/node_modules/playwright/lib/common/testType.js:225:17)
      at /Users/mm/p/livestore/examples/web-todomvc-sync-electric/tests/todomvc.spec.ts:3:6
      at loadAndEvaluateModule (2:1)
-------------------------------


 0 pass
 1 fail
 1 error
Ran 1 test across 1 file. [230.00ms]
```

Will have to see if switching back to `pnpm` helps

schickling:
Thanks a lot for opening this PR and helping to fix this. I generally would prefer using Bun as much as possible throughout the project. So I'm wondering whether there is an underlying Bun issue that we could reference here?

markmiro:
Thanks for getting back to me on this!

Bun was more brittle. I could get `bun dev` working though. My initial approach was to use `bun`, but it has three issues:
1. `bun test` fails and I haven't found a fix
1. Bun is probably breaking from a minor version bump to `"@tanstack/react-router": "^1.132.47"`. Setting to exact version (without caret) works.
1. Running `bun i` hangs at:
    ```
    bun install v1.3.3 (274e01c7)
       🔍 @effect/workflow [214/213] 
    ```
    Have to kill the process and run it again for it to succeed

`pnpm` doesn't have any of the above issues. `bun test` could probably be made to work, but for now I'm happy that I have a working example that I can use for the main project I'm contributing to.

markmiro:
@schickling just got it working with `bun` now too

schickling:
Thanks a lot. Appreciate the extra effort.

PR #867, opened 2025-11-29
Restore React KeyboardEvent in UI state tutorial

schickling:
## Summary
- switch the Chapter 6 tutorial examples back to using React.KeyboardEvent for the keydown handler
- align the import snippets and final code sample with the React namespaced type import

## Testing
- direnv exec . mono lint --fix *(fails: `mono` command not found in env)*

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_692ae52280288329b7b2b4eda7a1f436)


schickling:
@nikolasburk Would you mind giving this a quick review?

docs/src/content/docs/tutorial/6-persist-ui-state.mdx
```diff
@@ -288,10 +291,12 @@ const todos$ = queryDb((
 const todos = store.useQuery(todos$)
 ```
 
+Here, `uiState$` comes from `useClientDocument`, ensuring that the todos query automatically reacts whenever the persisted UI state changes.
+
 This is the final code for the `App` component:
 
 ```tsx title="src/App.tsx"
-import { useState } from 'react'
+import type React from 'react'
```
nikolasburk: this may not necessary: 

image

nikolasburk (approved):
looks great, thanks! I tested the change manually and it works 🙏  only small note is that the `React` import doesn't seem to be necessary, so we could consider dropping it. not a blocker for merging though.

PR #878, opened 2025-12-01
fix(ci): `build-example-create` failing on PR refs

### LINKED ISSUE #871: build-example-create fails on PR refs without full prefix
### Bug Description
`build-example-create` jobs on PRs exit early because the CLI is called with `--branch ${{ github.ref_name }}` (e.g. `865/merge`). GitHub's public Contents/Tarball APIs don't resolve shortened merge refs, so the scaffolding step gets a 404 when fetching the example source.

### Minimal Reproduction
1. Open any PR (e.g. #865) and wait for `ci / build-example-create` to run.
2. The job runs `pnpm dlx @livestore/cli@0.0.0-snapshot-${{ github.sha }} create --example <example> --branch ${{ github.ref_name }}`.
3. For PRs this passes `865/merge`; the CLI hits the GitHub Contents API with that ref and receives 404.
4. Job fails ~30s in with a NetworkError.

Minimal reproduction outside CI:
- `curl -s https://api.github.com/repos/livestorejs/livestore/contents/examples?ref=865/merge` → 404
- `curl -s https://api.github.com/repos/livestorejs/livestore/contents/examples?ref=refs/pull/865/merge` → OK

### LiveStore Version
0.0.0-snapshot-${{ github.sha }} (CI)

### Affected Packages
@livestore/cli; examples/web-todomvc, web-linearlite, expo-linearlite

### Environment
- OS: GitHub Actions runner (`namespace-profile-linux-x86-64`)
- Node.js version: runner default (pnpm/action-setup@v4 with standalone pnpm)
- Package manager: pnpm
- Runtime: GitHub Actions

### Expected Behavior
The example scaffold step should fetch sources for the PR branch/merge ref and complete so the example builds succeed.

### Actual Behavior
The CLI call using `--branch ${{ github.ref_name }}` passes `865/merge` to GitHub's API, which returns 404 (`No commit found for the ref 865/merge`), causing the job to fail/cancel.

### Error Logs
NetworkError: Failed to parse GitHub API response: Expected ReadonlyArray<...>, actual {"message":"No commit found for the ref 865/merge","documentation_url":"https://docs.github.com/rest/repos/contents#get-repository-content","status":"404"}

### Additional Context
Root cause: GitHub external APIs require full refs or branch/sha; `github.ref_name` (`865/merge`) is only understood inside Actions.
Fix options:
- Use `github.ref` (full `refs/pull/865/merge`) so the CLI tests the merge commit.
- Or use `github.head_ref`/`env.GITHUB_BRANCH_NAME` so the CLI uses the PR branch name; simplest because the env var already exists.
Which semantic do we want for this job? Testing merged state → `github.ref`; testing PR branch as-is → `github.head_ref`.

Checklist:
- [x] Searched existing issues
- [x] Included minimal reproduction
- [x] Included requested details

IGassmann:
## Summary

- Use `github.ref` instead of `github.ref_name` in CI because GitHub's public API requires full refs (e.g., `refs/pull/123/merge` not `123/merge`)
- Rename CLI `--branch` option to `--ref` to align with GitHub API parameter naming, since the value can be a branch, tag, commit SHA, or full ref
- Add `--branch`, `--tag`, `--commit` as aliases
- Fix tarball path for refs containing slashes (e.g., `refs/pull/123/merge`) by using `sluggify` to sanitize the filename

## Test plan

- [x] Verify `build-example-create` job passes on this PR
- [x] Test CLI with `--ref` and `--branch` options
- [x] Test CLI with full PR ref format (`refs/pull/878/head`)

Fixes #871

### DIFF
--- a/packages/@livestore/cli/src/commands/new-project.ts
+++ b/packages/@livestore/cli/src/commands/new-project.ts
@@ -40,11 +40,11 @@ export class DirectoryExistsError extends Schema.TaggedError<DirectoryExistsErro
 }) {}
 
 // Fetch available examples from GitHub
-const fetchExamples = (branch: string) =>
+const fetchExamples = (ref: string) =>
   Effect.gen(function* () {
-    const url = `https://api.github.com/repos/livestorejs/livestore/contents/examples?ref=${branch}`
+    const url = `https://api.github.com/repos/livestorejs/livestore/contents/examples?ref=${ref}`
 
-    yield* Effect.log(`Fetching examples from branch: ${branch}`)
+    yield* Effect.log(`Fetching examples from ref: ${ref}`)
 
     const request = HttpClientRequest.get(url)
     const response = yield* HttpClient.execute(request).pipe(
@@ -100,13 +100,13 @@ const selectExample = (examples: string[]) =>
   })
 
 // Download and extract example using tiged approach
-const downloadExample = (exampleName: string, branch: string, destinationPath: string) =>
+const downloadExample = (exampleName: string, ref: string, destinationPath: string) =>
   Effect.gen(function* () {
-    yield* Console.log(`📥 Downloading example "${exampleName}" from branch "${branch}"...`)
+    yield* Console.log(`📥 Downloading example "${exampleName}" from ref "${ref}"...`)
 
     const tempDir = yield* Effect.sync(() => os.tmpdir())
-    const tarballPath = nodePath.join(tempDir, `livestore-${branch}-${Date.now()}.tar.gz`)
-    const tarballUrl = `https://api.github.com/repos/livestorejs/livestore/tarball/${branch}`
+    const tarballPath = nodePath.join(tempDir, `livestore-${ref}-${Date.now()}.tar.gz`)
+    const tarballUrl = `https://api.github.com/repos/livestorejs/livestore/tarball/${ref}`
 
     // Download tarball directly
     const request = HttpClientRequest.get(tarballUrl)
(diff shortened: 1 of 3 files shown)


packages/@livestore/cli/src/commands/new-project.ts
```diff
@@ -206,9 +200,14 @@ export const createCommand = Cli.Command.make(
       Cli.Options.optional,
       Cli.Options.withDescription('Example name to create (bypasses interactive selection)'),
     ),
-    branch: Cli.Options.text('branch').pipe(
+    ref: Cli.Options.text('ref').pipe(
+      Cli.Options.withAlias('commit'),
```
schickling, line 204: I don't think we need all of those aliases, but otherwise looks good to me.

schickling (changes requested):

packages/@livestore/cli/src/commands/new-project.ts
IGassmann: I believe the aliases are useful because many developers aren't well familiar with git refs. This can make the following command confusing (e.g. when displayed in some docs or logs) for many:

```shell
pnpx @livestore/cli create --example web-todomvc --ref experiment-multi-store-api
pnpx @livestore/cli create --example web-todomvc --ref 664b708
```

Compared with:

```shell
pnpx @livestore/cli create --example web-todomvc --branch experiment-multi-store-api
pnpx @livestore/cli create --example web-todomvc --commit 664b708
```

What do you think?

schickling (approved):

PR #881, opened 2025-12-01
feat(ci): add run ID affinity to Namespace runners for faster workflow runs

IGassmann:
## Summary

- Add `namespace-features:github.run-id` label to all Namespace runner jobs to create workflow run affinity
- This ensures runners spawned for a workflow run can only execute jobs from that same run
- Fixes job scheduling issue where newer workflow runs would "jump the queue" ahead of older runs when hitting concurrency limits, causing longer workflow run times

## Problem

By default, any Namespace runner with matching labels can pick up any waiting job, even from different workflow runs. When hitting concurrency limits, this causes jobs from newer runs to be scheduled before jobs from older, incomplete workflow runs. The result is that older workflows take significantly longer to complete.

## Solution

Adding `namespace-features:github.run-id=${{ github.run_id }}` as a runner label creates affinity between runners and their workflow run. Since Namespace spawns runners in the order it learns about jobs, older workflow runs will have their runners assigned first and complete before newer runs consume runner capacity.

## Test plan

- [x] Verify CI workflow syntax is valid
- [ ] Monitor workflow execution times after merging to confirm improved scheduling behavior

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,15 @@ env:
 
 jobs:
   lint:
-    runs-on: namespace-profile-linux-x86-64
+    runs-on:
+      - namespace-profile-linux-x86-64
+      # By default, any runner with matching labels can pick up any waiting job, even from different
+      # workflow runs. This causes jobs from newer runs to "jump the queue" ahead of older runs when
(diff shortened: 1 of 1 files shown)


schickling (approved):
I hate yaml... 😂

PR #920, opened 2025-12-10
feat(livestore): add store lifecycle hooks for external state sync

colelawrence:
## Problem

LiveStore lacks lifecycle hooks that external state management systems (Jotai, Preact Signals, etc.) need to safely synchronize their state with LiveStore's reactive graph.

When integrating LiveStore with external reactive systems that have their own batching mechanisms, we need guarantees that:
1. We can observe when changes are about to happen (to capture old state if needed)
2. We can observe when the reactive graph has fully settled (safe to sync)
3. Any state changes triggered during sync callbacks start fresh cycles (batch isolation)

Without these hooks, integrations must resort to fragile workarounds or risk race conditions where external reactive updates get captured in the wrong LiveStore batch.

## Solution

Add two store lifecycle hooks accessible via `store[StoreInternalsSymbol]` (progressive disclosure—keeps main store API clean for typical users):

- **`subscribeToBeforeChange(callback)`** - Fires before `commit()` or `setSignal()` applies changes. The callback receives a discriminated union (`BeforeChangeEvent`) indicating whether it's a commit (with events) or setSignal (with signal def and new value).

- **`subscribeToAfterRefresh(callback)`** - Fires after the reactive graph has completely settled. The callback receives `AfterRefreshInfo` with the refresh reason, which atoms were recomputed, and timing info.

Key design decisions:
- `onAfterRefresh` fires **inside** the `effectsWrapper` (React's `batchUpdates`) but **after** all effects have run, ensuring the batch is truly closed
- Changes made from within `onAfterRefresh` trigger new, separate refresh cycles (batch isolation guarantee)
- The `BeforeChangeEvent` discriminated union allows type-safe handling of commits vs signal updates
- APIs are on `StoreInternals` rather than top-level store to honor progressive disclosure

Trade-off: We chose to fire `onAfterRefresh` per-refresh-cycle rather than debouncing, giving integrations precise control over sync timing at the cost of potentially more callback invocations.

## Validation

Added 14 unit tests covering:
- Basic subscription/unsubscribe for both hooks
- Discriminated union handling for commits and setSignal
- Updater function values resolved before `onBeforeChange` fires
- **Batch isolation**: `setSignal` and `commit` from within `onAfterRefresh` trigger NEW refresh cycles
- **Lifecycle order**: `onBeforeChange` (old value) → mutation → effects → `onAfterRefresh` (new value)
- `skipRefresh` + `manualRefresh()` behavior

```
✓ src/store/store-hooks.test.ts (14 tests) 257ms
Test Files 1 passed (1)
Tests 14 passed (14)
```

Also verified existing signal and db-query tests still pass.

### Demo

Lifecycle order test output showing the exact sequence:

```typescript
store.setSignal(value$, 'updated')

// Timeline captured:
expect(timeline).toEqual([
  'effect:initial',      // subscription setup
  'before:initial',      // onBeforeChange sees OLD value
  'effect:updated',      // reactive effect fires with NEW value  
  'after:updated'        // onAfterRefresh - batch closed, safe to sync
])
```

Batch isolation test - counter increments from callback trigger separate cycles:

```typescript
// Each increment in onAfterRefresh triggers a NEW refresh cycle
expect(refreshCycles).toEqual([1, 2, 3])  // 3 distinct cycles, not batched
```

### DIFF
--- a/packages/@livestore/livestore/src/reactive.ts
+++ b/packages/@livestore/livestore/src/reactive.ts
@@ -210,6 +210,7 @@ export class ReactiveGraph<
   private deferredEffects: Map<Effect<TDebugRefreshReason>, Set<TDebugRefreshReason>> = new Map()
 
   private refreshCallbacks: Set<() => void> = new Set()
+  private afterRefreshCallbacks: Set<(info: RefreshDebugInfo<TDebugRefreshReason, TDebugThunkInfo>) => void> = new Set()
 
   private nodeIdCounter = 0
   private uniqueNodeId = () => `node-${++this.nodeIdCounter}`
@@ -507,6 +508,7 @@ export class ReactiveGraph<
       this.debugRefreshInfos.push(refreshDebugInfo)
 
       this.runRefreshCallbacks()
+      this.runAfterRefreshCallbacks(refreshDebugInfo)
     })
   }
 
@@ -529,6 +531,12 @@ export class ReactiveGraph<
   runRefreshCallbacks = () => {
     for (const cb of this.refreshCallbacks) {
       cb()
+    }
+  }
+
+  private runAfterRefreshCallbacks = (info: RefreshDebugInfo<TDebugRefreshReason, TDebugThunkInfo>) => {
+    for (const cb of this.afterRefreshCallbacks) {
+      cb(info)
     }
   }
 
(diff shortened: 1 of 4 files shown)


schickling:
Thanks a lot for this PR. Would you mind taking a look at the failing docs build step as well? Seems to be related to some type checking.

PR #936, opened 2025-12-15
[Expo adapter]: Add configuration option for database directory

mikekeehnen:
## Problem

Expo SQLite only allowed database files to be stored in the default SQLite directory, preventing users from customizing the
storage location for their application databases.

## Solution

Added an optional directory configuration parameter to MakeDbOptions.storage that allows specifying a custom base directory for
database files. The implementation falls back to SQLite.defaultDatabaseDirectory when not specified, maintaining backward
compatibility.

## Validation

Not run - this is a straightforward API addition that extends existing configuration options without modifying default behavior.

## Related issues

None


schickling:
Thanks a lot for this PR. I've added a few documentation improvements. This will be released shortly.

PR #947, opened 2025-12-17
feat(react)!: remove top-level useQuery and useClientDocument exports

### LINKED ISSUE #946: refactor(react): remove top-level useQuery and useClientDocument exports
## Summary

Follow-up from #922 (multi-store API promotion). The `useQuery`, `useQueryRef`, and `useClientDocument` hooks are still exported at the top level from `@livestore/react`, but they should only be available as store methods (`store.useQuery()`, `store.useClientDocument()`).

## Current State

In `packages/@livestore/react/src/mod.ts`:
```ts
export {
  type UseClientDocumentResult,
  useClientDocument,
} from './useClientDocument.ts'
export { useQuery, useQueryRef } from './useQuery.ts'
```

This allows:
```ts
import { useQuery, useClientDocument } from '@livestore/react'
```

## Expected State

These hooks should only be accessible via the store instance:
```ts
const store = useAppStore()
const todos = store.useQuery(visibleTodos$)
const [doc, setDoc] = store.useClientDocument(tables.uiState)
```

## Tasks

- [ ] Remove top-level exports of `useQuery`, `useQueryRef`, and `useClientDocument` from `mod.ts`
- [ ] Verify all examples and docs use `store.useQuery()` / `store.useClientDocument()` pattern
- [ ] Update the migration guide in CHANGELOG.md to clarify that `useQuery()` and `useClientDocument()` are no longer available as standalone imports

## Related

- Parent PR: #922
- Issue: #841

IGassmann:
Closes #946

## Summary

- Removes top-level exports of `useQuery`, `useQueryRef`, and `useClientDocument` from `@livestore/react`
- Users must now access these hooks via the store instance: `store.useQuery()`, `store.useClientDocument()`
- Type exports (`UseClientDocumentResult`, `Dispatch`, `SetStateAction`, etc.) remain available at the top level

## Migration

```typescript
// Before
import { useQuery, useClientDocument } from '@livestore/react'
const todos = useQuery(query$)
const [state, setState] = useClientDocument(table)

// After
import { useStore } from '@livestore/react'

const store = useStore(storeOptions) // or via a custom hook wrapping useStore() (e.g. useAppStore())
const todos = store.useQuery(query$)
const [state, setState] = store.useClientDocument(table)
```

## Rationale

This change enforces a cleaner API pattern where queries are always explicitly associated with a store instance, removing ambiguity about which store is being queried.

## Test plan

- [x] TypeScript compilation passes
- [x] Linting passes
- [x] Documentation snippets updated


acusti (commented):
the motivation behind this PR makes sense to me, but i don't get the specific examples. where is `useAppStore` coming from? i only see `declare const useAppStore: ...`

am unimportant aside: "App Store" is a loaded term. but i think it makes sense here, so 🤷

IGassmann:
> the motivation behind this PR makes sense to me, but i don't get the specific examples. where is `useAppStore` coming from? i only see `declare const useAppStore: ...`
> 
> am unimportant aside: "App Store" is a loaded term. but i think it makes sense here, so 🤷

@acusti Thanks for the feedback! I've updated the snippets to import `useAppStore` from a shared store file (`reference/framework-integrations/react/store.ts`) that shows the actual implementation:

```typescript
export const useAppStore = () =>
  useStore({
    storeId: 'app-root',
    schema,
    adapter,
    batchUpdates,
  })
```

The migration example in the CHANGELOG and PR description now also hints at this pattern with `// or via a custom hook wrapping useStore() (e.g. useAppStore())`.

docs/src/content/_assets/code/reference/framework-integrations/react/schema.ts
```diff
@@ -7,25 +7,29 @@ export const tables = {
       id: State.SQLite.text({ primaryKey: true }),
       text: State.SQLite.text(),
       completed: State.SQLite.boolean({ default: false }),
+      createdAt: State.SQLite.datetime(),
```
acusti, line 10: good call on adding `createdAt` to the reference

docs/src/content/_assets/code/reference/reactivity-system/react-component.tsx
```diff
@@ -1,11 +1,12 @@
...
-import { useQuery } from '@livestore/react'
 import type { FC } from 'react'
+import { useAppStore } from '../framework-integrations/react/store.ts'
 
 declare const state$: LiveQueryDef<number>
 
 export const MyComponent: FC = () => {
-  const value = useQuery(state$)
+  const store = useAppStore()
+  const value = store.useQuery(state$)
```
acusti: this is out of scope (pre-existing and largely unrelated to removing the top-level `useQuery` and `useClientDocument` exports), but this is another code example where the intention isn’t clear to me on what a variable refers to or how i should use it in my own codebase (in this case `state$`)

CHANGELOG.md
```diff
@@ -219,6 +219,24 @@
...
+  ```typescript
+  // Before
+  import { useQuery, useClientDocument } from '@livestore/react'
+  const todos = useQuery(query$)
+  const [state, setState] = useClientDocument(table)
+
+  // After
+  import { useStore } from '@livestore/react'
+
+  const store = useStore(storeOptions) // or via a custom hook wrapping useStore() (e.g. useAppStore())
```
acusti, line 233: 👍

acusti (commented):
this looks good to me! there’s a handful of pnpm lockfile changes (primarily a new dependency on `astro@5.16.4` and a new optional dependency on `expo-router@6.0.10`), so hopefully those are expected, but definitely those docs changes look clear and correct.

acusti:
@IGassmann i submitted my approving review as a comment, but happy to submit an `Approve` review if that would be better, i just didn’t want to overstep

docs/src/content/_assets/code/reference/reactivity-system/react-component.tsx
IGassmann: Thanks! Addressed in 9ab467cba19c592cee4f32fa99faa95f9682f130

schickling (approved):

PR #953, opened 2025-12-23
[takeover required] misc sync backend fixes

Sam-Kruglov:
## Problem

I'm making a custom sync backend via a pull, a push, and a ping endpoints without long-polling. 

I've had multiple issues with the setup that this PR fixes, however I do need you to take it over as I'm not very well versed here and effect and didn't add any tests (haven't tried writing tests with effect yet). I added multiple FIXME comments for notable issues.

I've copy-pasted the webmvc example into my app for quick testing. In general I tested multiple scenarios where the backend was offline, then came online, was behind local and was ahead of local and both behind & ahead. Multiple issues fixed, I'll just comment in-code

https://discord.com/channels/1154415661842452532/1452545467316568135


packages/@livestore/common/src/index.ts
```diff
@@ -15,3 +15,4 @@ export * from './sync/index.ts'
 export * as SyncState from './sync/syncstate.ts'
 export * from './util.ts'
 export * from './version.ts'
+export * from './errors.js'
```
Sam-Kruglov, line 18: Since my sync backend is outside of the livestore packages (for now at least, the backend I'm connecting to is open source, but my implementation is not really production ready). You may further expand on this, e.g. `PullResItem` is also not visible

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -264,9 +258,22 @@ export const makeLeaderSyncProcessor = ({
...
       const span = yield* Effect.currentSpan.pipe(Effect.orDie)
       const otelSpan = yield* OtelTracer.currentOtelSpan.pipe(Effect.catchAll(() => Effect.succeed(undefined)))
-      const { devtools, shutdownChannel } = yield* LeaderThreadCtx
+      const { syncBackend, devtools, shutdownChannel } = yield* LeaderThreadCtx
       const runtime = yield* Effect.runtime<LeaderThreadCtx>()
 
+      // FIXME: validate not undefined
+      //   configure ping interval
+      //   somehow infer syncBackend.connect from it?
+      yield* syncBackend!.ping.pipe(
```
Sam-Kruglov, line 267: I can define the singular ping and I think LS should handle the rest as it sees fit. However, IMO we don't even need the ping given we periodically pull anyway. Perhaps also make it optional, same for the `connect`

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -275,6 +282,7 @@ export const makeLeaderSyncProcessor = ({
       }
 
       /** State transitions need to happen atomically, so we use a Ref to track the state */
+      yield* Effect.logDebug(`Init syncState: local=${initialSyncState.localHead.global}, upstream=${initialSyncState.upstreamHead.global}`)
```
Sam-Kruglov, line 285: I added several `syncState` debug logs whenever the sequence numbers for local/remote change, it's not too much of logs IMO and it's on debug level but feel free to adjust. It was crucial during validating this

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -303,15 +311,15 @@ export const makeLeaderSyncProcessor = ({
...
-            if (LS_DEV) {
-              yield* Effect.logDebug(
-                `Ignoring sync error (${cause._tag === 'Fail' ? cause.error._tag : cause._tag})`,
-                Cause.pretty(cause),
-              )
-            }
+            yield* Effect.logDebug(
+              `Ignoring sync error (${cause._tag === 'Fail' ? cause.error._tag : cause._tag})`,
+              Cause.pretty(cause),
+            )
```
Sam-Kruglov, line 317: IMO ignoring the error here may not be the best approach, I was having trouble while declaring the sync backend, I was getting lots of errors on my end and I didn't see much - e.g. I defined the response schema wrong and it was throwing parse error wrapped into unknown error and no logs. Same for the below thread crash. 

Besides logging, not sure why would we ignore this because lets say my reponse schema is wrong, which is unrecoverable, so e.g. backend pulling process will permanently crash and only page refresh will try it again. I would just let it crash completely and log error.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -336,6 +344,9 @@ export const makeLeaderSyncProcessor = ({
 
       const backendPushingFiberHandle = yield* FiberHandle.make<void, never>()
       const backendPushingEffect = backgroundBackendPushing({
+        pullLatch,
+        localPushesLatch,
+        syncStateSref,
```
Sam-Kruglov, line 349: I'm reusing local pushes latch here, I think should be ok, perhaps should rename it to just `pushesLatch`. Awaiting on both latches inside.

By the way, I find it a bit weird that pulling locks pushing and pushing locks pulling. Perhaps introduce a third latch like `sync` so that only one process can use that at a time. But there's also ordering to think about like `1) pull 2) local push 2) remote push`. Perhaps introducing a separate class to manage that would be cleaner. Not a high priority though

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -370,8 +381,12 @@ export const makeLeaderSyncProcessor = ({
         advancePushHead,
       }).pipe(
         Effect.retry({
+          // FIXME: shouldn't maybeShutdownOnError decide whether we shut down or not?
+          //  This will stop retrying if while returns false and require page refresh to reconnect
           // We want to retry pulling if we've lost connection to the sync backend
           while: (cause) => cause._tag === 'IsOfflineError',
+          // FIXME: make that configurable, should match the non-failure repeat inside
+          schedule: Schedule.spaced("10 seconds")
```
Sam-Kruglov, line 389: this is the case I was talking about above when leader thread was half-crashing without logs

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
     delay: Effect.Effect<void> | undefined
   }
 }) =>
-  Effect.gen(function* () {
-    while (true) {
-      if (testing.delay !== undefined) {
-        yield* testing.delay.pipe(Effect.withSpan('localPushProcessingDelay'))
-      }
+  Effect.scoped(Effect.gen(function* () {
```
Sam-Kruglov, line 472: I re-wrote `while (true)` into `Effect.repeat`

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
-      // Wait for the backend pulling to finish
-      yield* localPushesLatch.await
+    // Wait for the backend pulling to finish
+    yield* localPushesLatch.await
 
-      // Prevent backend pull processing until this local push is finished
-      yield* pullLatch.close
+    // Prevent backend pull processing until this local push is finished
+    yield* pullLatch.close
+    yield* Effect.addFinalizer(() => pullLatch.open)
```
Sam-Kruglov, line 484: this is **CRUCIAL**,  there was a bug where an error was thrown below (due to backend being offline) and the latch remained closed forever. I'm not sure if we need to wrap this into `Effect.sync` but I think this works. I saw other uses of `addFinalizer` so I'm guessing you know about this but make sure there aren't any other places in the codebase that release resources manually.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
+        `push:drop-old-generation`,
+        {
+          droppedCount: droppedItems.length,
+          currentRebaseGeneration,
+        },
+        undefined,
       )
 
-      if (droppedItems.length > 0) {
+      /**
```
Sam-Kruglov, line 508: this is copy-pasted unchanged from below

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
-              newEvents: TRACE_VERBOSE ? JSON.stringify(newEvents) : undefined,
-            },
-            undefined,
-          )
-          return yield* new UnknownError({ cause: mergeResult.message })
+        // TODO we still need to better understand and handle this scenario
+        if (LS_DEV && (yield* BucketQueue.size(localPushesQueue)) > 0) {
+          console.log('localPushesQueue is not empty', yield* BucketQueue.size(localPushesQueue))
+          // biome-ignore lint/suspicious/noDebugger: debugging
+          debugger
```
Sam-Kruglov, line 583: still unchanged

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
-              deferred,
-              LeaderAheadError.make({ minimumExpectedNum: mergeResult.expectedMinimumId, providedNum }),
-            ),
-          )
+        yield* Effect.forEach(allDeferredsToReject, (deferred) =>
+          Deferred.fail(
+            deferred,
+            LeaderAheadError.make({ minimumExpectedNum: mergeResult.expectedMinimumId, providedNum }),
+          ),
+        )
```
Sam-Kruglov, line 596: still unchanged

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
 
-          // Allow the backend pulling to start
-          yield* pullLatch.open
+        // Allow the backend pulling to start
+        yield* pullLatch.open
```
Sam-Kruglov, line 599: actually since we have finalizer now, this can be removed, line 529 as well.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
-
-      yield* connectedClientSessionPullQueues.offer({
-        payload: SyncState.PayloadUpstreamAdvance.make({ newEvents: mergeResult.newEvents }),
-        leaderHead: mergeResult.newSyncState.localHead,
-      })
+    const localSeq = mergeResult.newSyncState.localHead.global
+    const upstreamSeq = mergeResult.newSyncState.upstreamHead.global
+    if (syncState.localHead.global != localSeq) {
+      yield* Effect.logDebug(`Updating syncState after local push: local=${localSeq}, upstream=${upstreamSeq}`)
+    }
```
Sam-Kruglov, line 617: only added the log here

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -454,177 +469,181 @@ const backgroundApplyLocalPushes = ({
...
 
-      // Allow the backend pulling to start
-      yield* pullLatch.open
-    }
-  })
+    yield* materializeEventsBatch({ batchItems: mergeResult.newEvents, deferreds })
+  })).pipe(
+    Effect.repeat(Schedule.forever),
+    Effect.withSpan('@livestore/common:LeaderSyncProcessor:local-pushing')
+  )
```
Sam-Kruglov, line 646: unchanged

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -709,26 +728,44 @@ const backgroundBackendPulling = ({
...
-      newEvents: LiveStoreEvent.Client.EncodedWithMeta[],
-      pageInfo: SyncBackend.PullResPageInfo,
-    ) =>
-      Effect.gen(function* () {
-        if (newEvents.length === 0) return
-
-        if (devtoolsLatch !== undefined) {
-          yield* devtoolsLatch.await
-        }
+    yield* Effect.scoped(Effect.gen(function* () {
```
Sam-Kruglov, line 731: scope for the finalizer for the latch; inline the `onNewPullChunk`

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -826,73 +863,45 @@ const backgroundBackendPulling = ({
...
+      }
+      yield* initialBlockingSyncContext.update({ processed: batch.length, pageInfo })
+    })).pipe(Effect.repeat(Schedule.spaced("10 seconds")))
     // Should only ever happen when livePull is false
     yield* Effect.logDebug('backend-pulling finished', { livePull })
   }).pipe(Effect.withSpan('@livestore/common:LeaderSyncProcessor:backend-pulling'))
 
 const backgroundBackendPushing = ({
+  pullLatch,
+  localPushesLatch,
```
Sam-Kruglov, line 881: await on these 2, mentioned above

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -958,10 +966,21 @@ const backgroundBackendPushing = ({
...
+            yield* Effect.logDebug('handled unknown backend-push-error', { error })
           }
 
           return yield* error
         }
+        const newBackendHead = queueItems.at(-1)!.seqNum
+        Eventlog.updateBackendHead(dbEventlog, newBackendHead)
+        const syncState = (yield* syncStateSref)!
+        // @ts-ignore readonly
+        syncState.upstreamHead = newBackendHead
```
Sam-Kruglov, line 979: This is also **MAJOR**, the remote head was never updated after backend push.

there's probably a cleaner way to update this via merge? 

I was a bit confused about backend head separately vs sync state which appears to have that inside but whenever I only update the head, it is still outdated inside the sync state. I suggest removing the separate backend head methods and funnel it through sync state so that there's less confusion and bug probability

packages/@livestore/common/src/sync/sync-backend.ts
```diff
@@ -61,6 +61,21 @@ export type SyncBackend<TSyncMetadata = Schema.JsonValue> = {
...
+    }>,
+    options?: {
+      /**
+       * If true, the sync backend will return a stream of events that have been pushed after the cursor.
+       *
+       * @default false
+       */
+      live?: boolean
+    },
+  ) => Effect.Effect<PullResItem<TSyncMetadata>, IsOfflineError | InvalidPullError>
```
Sam-Kruglov, line 78: This is also major and a breaking change. My suggestion is to make all ping/connect/push/pull an effect. Currently pull is a stream but I noticed that the periodic execution only  happens inside the stream, which means the cursor is not updated - it takes the cursor info on init and from then doesn't update it. Specifically, it is not updated after a backend push, but it is updated update a pull which contains new events. 

I introduced a new optional field not to break anything but you can see that in the code I'm now only referring to it like `syncBackend.pullEffect!(...`, so be careful here. If you accept, the intention is to adjust the existing sync backends and change the return type of `pull`.

Anyway, by making that into an effect, livestore is now responsible for handling the periodic pulling, i.e. the scheduling and error handling. Could be good to make the intervals/timeouts configurable. 

**Most importantly**, the cursor is now updated on every pull. Together with the comment right above this one where the backend push now actually updates the upstream head number, my stuff (end-to-end frontend-to-backend event sourcing) finally works!

Thanks for this project! Please, take over this PR as I'm running low on time and also codebase/effect knowledge to do this properly and add tests...

Sam-Kruglov:
just saw your message in discord, replied [there](https://discord.com/channels/1154415661842452532/1452545467316568135/1452940292029288479)

branch is out of date but I just pulled yesterday and don't want comments to fall off

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
Sam-Kruglov: upd: we talked about how it's better to do a pull after push and dedup but in my case it just pushes the same event again. I'm looking for a way to make the push idempotent but livestore should be able to tell if the event that came from the pull already exists locally and NOT push it, right?

PR #968, opened 2026-01-08
fix(livestore): auto-dispose StoreRegistry runtime on garbage collection

### LINKED ISSUE #961: [Bug]: WS connection is not terminated when store unmounts
### Bug Description

When using LiveStore with Cloudflare sync the WS connection is not terminated after page with LiveStore component unmounts.

### Minimal Reproduction
Initially discovered this issue with Astro + React setup:
https://github.com/bohdanbirdie/astro-livestore-ws-issue

Confirmed in plain React SPA:
https://github.com/bohdanbirdie/livestore-ws-termination-issue

### LiveStore Version

0.4.0-dev.22

@livestore/adapter-web, @livestore/sync-cf

### Environment

```markdown
- OS: MacOS 15.6.1
- Node.js version: Bun 1.3.5 
- Package manager: Bun
- Runtime: Browser
```

### Expected Behavior

When component with livestore unmounts and store is no longer used the WS connection should terminate

### Actual Behavior

When component with livestore unmounts the WS connection retained

### Error Logs/Screenshots

https://github.com/user-attachments/assets/8518b748-8bbf-4806-b7d2-2cc5f78dc778

### Additional Context

https://discord.com/channels/1154415661842452532/1455536237292294228

### Checklist

IGassmann:
## Summary

- Use `FinalizationRegistry` to automatically dispose the `ManagedRuntime` when a `StoreRegistry` instance is garbage collected
- Ensures store resources (including WebSocket connections) are properly released when the registry goes out of scope

Fixes #961

## Problem

When a `StoreRegistry` goes out of scope (e.g., React component unmounts), JavaScript's garbage collector reclaims it, but this doesn't trigger Effect finalizers. This leaves WebSocket connections and other store resources open indefinitely.

## Solution

Register each `StoreRegistry` with a module-level `FinalizationRegistry`. When the registry is GC'd, the callback disposes the `ManagedRuntime`, which closes all store scopes and releases resources.

## Test plan

- [x] All existing `StoreRegistry` tests pass
- [ ] Manual verification with reproduction app: https://github.com/bohdanbirdie/livestore-ws-termination-issue

**Why no unit test for FinalizationRegistry:**

Direct testing of `FinalizationRegistry` behavior is not practical because:
- GC timing is non-deterministic and varies by environment/memory pressure
- `global.gc()` (with `--expose-gc` flag) doesn't guarantee finalization callbacks run immediately
- Per ECMAScript spec, finalization callbacks are not guaranteed to run at all

The `FinalizationRegistry` acts as a safety net for resource cleanup when the registry is garbage collected. Testing this would require forcing GC and observing side effects, which is inherently unreliable.

### DIFF
--- a/packages/@livestore/livestore/src/store/StoreRegistry.ts
+++ b/packages/@livestore/livestore/src/store/StoreRegistry.ts
@@ -187,6 +187,12 @@ export class StoreRegistry {
   readonly #defaultOptions: StoreRegistryConfig['defaultOptions']
 
   /**
+   * Function to dispose the managed runtime, if this registry owns it.
+   * Undefined when a custom runtime was provided (user owns cleanup).
+   */
+  readonly #disposeRuntime: (() => Promise<void>) | undefined
+
+  /**
    * Creates a new StoreRegistry instance.
    *
    * @example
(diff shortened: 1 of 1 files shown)


IGassmann:
## Progress Update

### What's done:
1. **Fixed GC reference cycle** - The `lookup` closure was capturing `this` via `this.#defaultOptions`, creating a reference path from Effect's global fiber roots back to the StoreRegistry. Fixed by copying `#defaultOptions` to a local variable before creating the RcMap.

2. **FinalizationRegistry now works** - StoreRegistry instances are now properly garbage collected, and the finalization callback fires to dispose the ManagedRuntime.

3. **Added `dispose()` method** - For cases where users create multiple short-lived registries (e.g., one per client island), they can call `dispose()` to immediately release resources and avoid lock conflicts with subsequent registries.

### Outstanding issue:
**`dispose()` doesn't work with React strict mode** - In strict mode, React runs effects twice (mount → cleanup → mount). Since `useState` preserves the same registry instance, the cleanup disposes a registry that's still being used on remount.

Will continue investigating solutions in a later day.

PR #991, opened 2026-01-25
Unify SQL `BindValues` type across codebase

TheAndrewJackson:
## Problem

SQL bind values were represented inconsistently across the codebase (multiple “bind value” types, mixed import sources, and frequent `as any as PreparedBindValues` casts). This made it difficult to reason about whether the `bindValues` used in the codebase actually needed different types.

```ts
//packages/@livestore/common/src/util.ts
export type SqlValue = string | number | Uint8Array<ArrayBuffer> | null
export type ParamsObject = Record<string, SqlValue>
export type Bindable = ReadonlyArray<SqlValue> | ParamsObject

//packages/@livestore/common/src/sql-queries/sql-queries.ts
export type BindValues = {
  readonly [columnName: string]: any
}
```

## Solution

This PR unifies bind value types and standardizes normalization at the execution boundary:

- Introduces canonical bind-value types in `@livestore/common/src/util.ts`:
  - `SqlValue` (SQLite primitives)
  - `SqlBindValue` (`SqlValue` + `SessionIdSymbol` sentinel)
  - `BindValues` (positional array or named object)
  - `SqlBindValueSchema` for schema validation/tooling
- Updates QueryBuilder to return `BindValues` from `.asSql()` and updates AST→SQL encoding to validate bind values via `SqlBindValueSchema`.
- Ensures QueryBuilder execution/select paths normalize bind values via `prepareBindValues(...)` (instead of relying on unsafe casts or call-site conventions).
- Propagates the unified `BindValues` type through LiveStore (`Store.query`, live queries), adapter-web SQL helpers, CLI MCP runtime/tool definitions, and the Svelte wrapper.

**Trade-offs / alternatives considered**
- Alternative: keep separate bind types per package/feature area (QueryBuilder vs sql-queries vs adapters) and add adapters between them. This was rejected in favor of a single shared contract to reduce friction and eliminate repeated conversion logic.
- `prepareBindValues` still uses simple string containment to detect used named params (existing behavior); this PR keeps that approach and focuses on unifying types and call sites first.

### Before / After

**Before: multiple bind types + unsafe casts**
```ts
// ad-hoc bind shapes across the codebase
type Bindable = ReadonlyArray<SqlValue> | Record<string, SqlValue>

// QueryBuilder execution often forced into PreparedBindValues
const { query, bindValues } = qb.asSql()
db.select(query, bindValues as any as PreparedBindValues)
```

**After: single `BindValues` type + normalization at the DB boundary**
```ts
// @livestore/common/src/util.ts
export type SqlValue = string | number | Uint8Array<ArrayBuffer> | null
export type SqlBindValue = SqlValue | typeof SessionIdSymbol
export type BindValues = ReadonlyArray<SqlBindValue> | Readonly<Record<string, SqlBindValue>>

export const prepareBindValues = (values: BindValues, statement: string): PreparedBindValues => { /* ... */ }

// sqlite-db-helper.ts
const { query, bindValues } = qb.asSql()
db.select(query, prepareBindValues(bindValues, query))
```

### DIFF
--- a/packages/@livestore/adapter-web/src/common/connection.ts
+++ b/packages/@livestore/adapter-web/src/common/connection.ts
@@ -1,11 +1,10 @@
 import type { PreparedBindValues, SqliteDb } from '@livestore/common'
-import { prepareBindValues, SqliteError } from '@livestore/common'
-import type { BindValues } from '@livestore/common/sql-queries'
+import { prepareQueryBindValues, SqliteError, type SqlQueryBindParams } from '@livestore/common'
 import type { WaSqlite } from '@livestore/sqlite-wasm'
 import { Effect } from '@livestore/utils/effect'
 
-export const execSql = (sqliteDb: SqliteDb, sql: string, bind: BindValues) => {
-  const bindValues = prepareBindValues(bind, sql)
+export const execSql = (sqliteDb: SqliteDb, sql: string, bind: SqlQueryBindParams) => {
+  const bindValues = prepareQueryBindValues(bind, sql)
   return Effect.try({
     try: () => sqliteDb.execute(sql, bindValues),
     catch: (cause) =>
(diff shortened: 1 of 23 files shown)


TheAndrewJackson:
This has gotten stale. Closing for now to focus on other work

PR #999, opened 2026-01-30
chore: update s2 docker image

infiniteregrets:
We had a recent change in the docker images we provide to bundle s2-lite with the cli and thus the change in the invocation updated in the PR


infiniteregrets:
@schickling this fixes the recent CI failures for s2 sync

schickling:
Thanks a lot!

PR #1002, opened 2026-02-04
Misc sync backend

Sam-Kruglov:
Miscellaneous sync backend fixes.

I made a new sync backend and I have ran into multiple issues, here are the fixes I've had to make. I feel I've reached a more-or-less stable state of my project now, so I want to offload these.

**There are 8 commits**, you can review one-by-one like so:
Screenshot 2026-02-03 at 11 37 01 PM


packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -281,6 +281,7 @@ export const makeLeaderSyncProcessor = ({
       }
 
       /** State transitions need to happen atomically, so we use a Ref to track the state */
+      yield* Effect.logDebug(`Init syncState: local=${initialSyncState.localHead.global}, upstream=${initialSyncState.upstreamHead.global}`)
```
Sam-Kruglov, line 292: it was very hard to see any problems without such logs. E.g. I saw head going up continuously without me committing events; I saw upstream head not moving due to deadlock, I saw local head not moving due to deadlock, etc

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -392,8 +392,12 @@ export const makeLeaderSyncProcessor = ({
         advancePushHead,
       }).pipe(
         Effect.retry({
+          // FIXME: shouldn't maybeShutdownOnError decide whether we shut down or not?
+          //  This will stop retrying if while returns false and require page refresh to reconnect
           // We want to retry pulling if we've lost connection to the sync backend
           while: (cause) => cause._tag === 'IsOfflineError',
+          // FIXME: make that configurable, should match the non-failure repeat inside
+          schedule: Schedule.spaced("10 seconds")
```
Sam-Kruglov: without schedule it would fire requests as fast as it could flooding the network tab while the backend was offline

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -951,8 +955,7 @@ const backgroundBackendPushing = ({
       // TODO(metrics): expose counters/gauges for retry attempts and queue health via devtools/metrics
 
       // Only retry for transient UnknownError cases
-      const isRetryable = (err: InvalidPushError | IsOfflineError) =>
-        err._tag === 'InvalidPushError' && err.cause._tag === 'LiveStore.UnknownError'
+      const isRetryable = (err: InvalidPushError | IsOfflineError) => err._tag === 'IsOfflineError'
```
Sam-Kruglov: if the push is invalid, then it's not recoverable, it would be a bug. Unknown error even more so. Offline however, is the one that needs to be retried. Had to debug to get here to find out why it won't retry

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -480,182 +480,181 @@ const backgroundApplyLocalPushes = ({
     delay: Effect.Effect<void> | undefined
   }
 }) =>
-  Effect.gen(function* () {
-    while (true) {
-      if (testing.delay !== undefined) {
-        yield* testing.delay.pipe(Effect.withSpan('localPushProcessingDelay'))
-      }
+  Effect.scoped(Effect.gen(function* () {
```
Sam-Kruglov, line 488: consider going through all of your code with AI to find all spots where you release resources like latches manually and use `addFinalizer` instead or at least `try--finally`. Had some cases where hit the deadlock because of these conditional releases in this file

Also, go through the codebase and add timeouts to effects whenever you are waiting on latches or whatever else. 

^ created an issue for this #1003

replaced while with effect repeat

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -480,182 +480,181 @@ const backgroundApplyLocalPushes = ({
...
+    // Wait for the backend pulling to finish
+    yield* localPushesLatch.await
 
-      // Prevent backend pull processing until this local push is finished
-      yield* pullLatch.close
+    // Prevent backend pull processing until this local push is finished
+    yield* pullLatch.close
+    yield* Effect.addFinalizer(() => Effect.gen(function* () {
+      if ((yield* BucketQueue.size(localPushesQueue)) === 0) {
+        yield* pullLatch.open
```
Sam-Kruglov: in case local push queue is filled up more than the batch size, we should keep pushing until it's done and not let backend sync do anything or it may cause weird sync state or some race conditions - don't remember but there was something happening, maybe not here exactly, but I did the same for other `BucketQueue.take`

packages/@livestore/common/src/sync/syncstate.ts
```diff
@@ -465,7 +465,7 @@ export const findDivergencePoint = ({
 
     if (divergencePointWithoutClientEvents === -1) return -1
 
-    const divergencePointEventSequenceNumber = existingEvents[divergencePointWithoutClientEvents]!.seqNum
+    const divergencePointEventSequenceNumber = filteredExistingEvents[divergencePointWithoutClientEvents]!.seqNum
```
Sam-Kruglov, line 471: I'm quite certain this is was a typo and I had some weird bug because of it I hunted down to here. We spent some effort right above calculating the filtered events and then not using it

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -743,7 +743,13 @@ const backgroundBackendPulling = ({
       newEvents: LiveStoreEvent.Client.EncodedWithMeta[],
       pageInfo: SyncBackend.PullResPageInfo,
     ) =>
-      Effect.gen(function* () {
+      Effect.scoped(Effect.gen(function* () {
+        yield* Effect.addFinalizer(() => Effect.gen(function* () {
```
Sam-Kruglov, line 745: same as above, using `addFinalizer` to release latches. There was a bug right below this `if (newEvents.length === 0) return` - latch is not open in this case causing a deadlock or something weird. With this, we're guaranteeing the latch is opened and my problem went away

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -179,6 +179,7 @@ export const makeLeaderSyncProcessor = ({
 
     const localPushesQueue = yield* BucketQueue.make<LocalPushQueueItem>()
     const localPushesLatch = yield* Effect.makeLatch(true)
+    const syncPushesLatch = yield* Effect.makeLatch(false)
```
Sam-Kruglov: right now backend push and backend pull happen independently which is wrong, I think the docs even say that push is supposed to only happen after pull. I've had some weird issues without this. I'm using release instead of open/close here

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -756,6 +762,17 @@ const backgroundBackendPulling = ({
           yield* devtoolsLatch.await
         }
 
+        // FIXME: close it BEFORE the actual pull is done. This point in time is AFTER response is fetched.
```
Sam-Kruglov: Stream.tap is not ideal as we only ever pause after the pull is already done. While we do stuff, the old pull response is just sitting here and after the pull latch is opened, we process that stale response and move forward with backend push - violation - backend push must only happen after a fresh pull. It doesn't happen very often so I didn't fix it, just a FIXME for you

packages/@livestore/common/src/schema/state/sqlite/column-def.ts
```diff
@@ -195,22 +195,23 @@ const getColumnForSchema = (schema: Schema.Schema.AnyNoContext, nullable = false
...
-  )
-
-  if (coreTypes.length === 0 || coreTypes.length === ast.types.length) {
-    return ast
-  }
-
-  if (coreTypes.length === 1) {
-    return coreTypes[0]!
+  let cur = ast
+  // could be wrapped twice like [null | [undefined | X] ]
```
Sam-Kruglov, line 200: To make optional columns work with livestore I have to add `NullOr` to every column:
```ts
const TxTableOptionalsS = S.partial(
  S.Struct({
    memo: S.NullOr(S.String),
    updatedOn: S.NullOr(S.String),
...
```
If I do the above, this change is required to properly unwrap the actual value as it's now wrapped twice into nullable types.

If I just use `S.partial` then `null` is not allowed and effect schema fails to parse the select return value. In addition to that,`column-def.ts:80` prints these warnings:
```
Field 'memo' has both null and undefined - treating | undefined as | null
Field 'updatedOn' has both null and undefined - treating | undefined as | null
```

---

Speaking of this, it's kind of annoying to deal with nullability like the above, I would just prefer to wrap a struct in `S.partial` and not have to use `NullOr`, or use `optional()` on a single column without having to define the options to add `null`.

* SQLLite when returning fields returns `null` values.
* Effect is used very heavily.
* Effect's idiomatic way of saying that something is "optional" normally means "missing". It will also work with `undefined` but the preferred way is the actual absence. 
  * Please read their take on [exactOptionalPropertyTypes](https://effect.website/docs/schema/introduction/#the-exactoptionalpropertytypes-option) 
  * IMO, in typescript, "missing" is the cleanest way to represent an optional value. `undefined` is the next best thing, I would try to avoid it altogether if possible. `null` would be intentional empty state, I'd use that to reset some value, or to store some react state that user can set/unset.

packages/@livestore/common/src/sync/ClientSessionSyncProcessor.ts
```diff
@@ -99,6 +99,10 @@ export const makeClientSessionSyncProcessor = ({
       pending: [],
     }),
   }
+  // FIXME: https://github.com/livestorejs/livestore/issues/970
+  Effect.runCallback(clientSession.leaderThread.syncState.get.pipe(Effect.andThen(leaderState => {
```
Sam-Kruglov, line 106: this client session sync processor is conflicting with leader sync processor. It lags behind the leader, doesn't seem to sync with it on its own. On startup I added this to sync with leader, and on pull I added below to kind of carefully sync with leader. Both changes solved my problems but AI was saying it's wrong to fix it that way but I still fail to understand what this file is even for if we already have a leader sync processor...

packages/@livestore/common/src/sync/ClientSessionSyncProcessor.ts
```diff
@@ -202,6 +210,7 @@ export const makeClientSessionSyncProcessor = ({
       const onBeforeUnload = (event: BeforeUnloadEvent) => {
         if (syncStateRef.current.pending.length > 0) {
           // Trigger the default browser dialog
+          console.error('client-session-sync-processor still got pending events: ', syncStateRef.current.pending)
```
Sam-Kruglov, line 220: if the syncing process is stuck, the webpage works fine but if you try to navigate or refresh you can a dialog saying data will be lost like when you're filling out a form. I tracked 1 case to here related to pending

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
Sam-Kruglov: justification in the discord thread I believe was that it fits pull via websocket. But perhaps it could be separated for regular http pulls. For regular http pull there should be a delay between them, by default it would send pull as fast as it could

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
IGassmann: I've replaced it with the `pushPullMutex` which allows for the same behavior with a single primitive.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
IGassmann: The intent makes sense, but there's an issue with the implementation: `IsOfflineError` is already caught by the `Effect.never` handler above (alongside `ServerAheadError`), so it never reaches the retry schedule. This meant nothing was actually being retried.

`UnknownError` inside `InvalidPushError` can represents transient server failures (500s, network glitches) — these are recoverable and should be retried with backoff. Reverted in 0d395ac

IGassmann:
Thanks again for the contribution, @Sam-Kruglov!

I'm going to close this PR because there have been several recent changes to the files it touches, which means this would need a non-trivial rebase. Rather than trying to untangle that, I think it would work better to split this into smaller, focused PRs: one per fix or closely related group of fixes. That will make each one easier to review, rebase, and merge independently.

Would you be open to re-opening these as separate PRs? I'm happy to help with your changes.

PR #1006, opened 2026-02-05
refactor: replace latches with mutex in LeaderSyncProcessor

IGassmann:
## Problem

`backgroundApplyLocalPushes` and `backgroundBackendPulling` must not modify sync state concurrently. Previously this was enforced by two cross-referenced latches (`localPushesLatch` and `pullLatch`) where each side controlled the other's gate — push closed `pullLatch` and awaited `localPushesLatch`, while pull closed `localPushesLatch` and awaited `pullLatch`. Understanding the mutual exclusion required mentally tracking both latches and how they interacted across the two sides. The correctness depended on the specific ordering of `close`/`await` calls, meaning swapping the order, forgetting to reopen one latch, or adding a new code path that missed one of the two operations could silently break the coordination or cause a deadlock.

## Solution

Replace both latches with a single `pushPullMutex` (`Effect.Semaphore(1)`). Each side calls `take(1)` before modifying sync state and `release(1)` when done — a single primitive with a straightforward acquire/release contract instead of two cross-referenced gates. The semaphore's built-in FIFO ordering provides the same anti-starvation guarantee that the previous preemptive `close` pattern provided — once a pull's `take` is queued, no subsequent push can jump ahead. The pull side still holds the permit across paginated pulls by only calling `release` on `NoMore`, preserving the existing behavior.

## Test plan

- [x] TypeScript build passes (`mono ts`)
- [x] Linting passes (`mono lint --fix`)
- [x] All 14 LeaderSyncProcessor tests pass, including the `concurrent pushes` test that exercises push/pull interleaving

### DIFF
--- a/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
+++ b/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
@@ -68,8 +68,8 @@ type LocalPushQueueItem = [
  *   - Maintains events in ascending order.
  *   - Uses `Deferred` objects to resolve/reject events based on application success.
  * - Processes events from the queue, applying events in batches.
- * - Controlled by a `Latch` to manage execution flow.
- * - The latch closes on pull receipt and re-opens post-pull completion.
+ * - Controlled by a mutex (`Semaphore(1)`) to ensure mutual exclusion between push and pull processing.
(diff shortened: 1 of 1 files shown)


schickling (approved):
Looks good to me overall. I'm wondering whether this change has any measurable performance difference?

PR #1012, opened 2026-02-07
Enhancement: Event log O(1) lookup

OrkhanAlikhanov:
Improved event log lookup in the leader thread to O(1) for faster startup.

## Problem

Initially described in [discord](https://discord.com/channels/1154415661842452532/1154415662874247191/1469453604460036290) but briefly, when there were lots of unsynced local event logs (like 20k), the expo app would start up very late (29 seconds).

## Solution

It turned out that most of the time was spent inside `getEventsSince` method because of nested loops. Using precomputed map for lookup instead of calling `sessionChangesetRowsDecoded.find` inside the `map` loop reduced the duration from `29` seconds to `8` seconds (like 70% faster). Still higher but that's good enough for a simple change. 

The other time consuming part is `LiveStoreEvent.Client.EncodedWithMeta.make` call. I believe it runs validation on each call contributing to slowness. I was able to squeeze it to 4 seconds my replacing `LiveStoreEvent.Client.EncodedWithMeta.make` calls with object construction with manual prototype setter but that becomes very cumbersome and unnecessary I'd say.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,6 +310,8 @@ See the [S2 sync provider docs](https://dev.docs.livestore.dev/reference/syncing
 
 #### Core Runtime & Storage
 
+- **Event log lookup optimization:** Improved event log lookup performance for large unsynced logs, speeding startup time ([#1011](https://github.com/livestorejs/livestore/pull/1011)).
+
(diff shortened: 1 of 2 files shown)


schickling (approved):

schickling:
Thanks!

PR #1021, opened 2026-02-13
Fix changeset_apply crash: use proper WASM function pointer for conflict handler

### LINKED ISSUE #998: [Bug]: wa-sqlite changeset_apply throws "function signature mismatch" during rebase rollback
### Bug Description

I've been encountering this bug often in my LiveStore apps. It seems to be most prevalent when syncing ClientDocuments with Schema.Record between tabs due to the high probability of conflicts but from my understanding wouldn't be limited to ClientDocument:

Rebase rollback applies inverted changesets and sqlite3.changeset_apply crashes with RuntimeError: function signature mismatch (Failed calling makeChangeset.apply). The failure occurs during conflict handling in wa‑sqlite because a JS callback is passed to WASM instead of a function pointer, so any rollback that invokes changeset_apply can fail under concurrent sync.

See more info and two possible solutions in:
https://github.com/slashv/livestore/blob/repro/changeset-apply-bug/repro/CHANGESET_APPLY_BUG.md

### Minimal Reproduction

The issue is timing dependent. Currently implemented as repro app with playwright test and triggers only when running in --headed mode. The tests only fail if the exact changeset error is raised.

https://github.com/slashv/livestore/tree/repro/changeset-apply-bug

```bash
cd repro
pnpm playwright test --headed
```

### LiveStore Version

0.4.0-dev.22

@livestore/wa-sqlite

### Environment

```markdown
- OS: MacOs
- Node.js version: 
- Package manager: pnpm
- Runtime: (browser/Node.js/Cloudflare Workers/other)
```

### Expected Behavior

The conflict resolution should resolve correctly and LiveStore should remain alive.

### Actual Behavior

Error is raised and LiveStore stops:
[@livestore/adapter-web:client-session] client-session shutdown

### Error Logs/Screenshots

```shell
Error message:
LiveStore.SqliteError: { "query": undefined, "code": -1, "cause": RuntimeError: function signature mismatch, "note": "Failed calling makeChangeset.apply" }
Cause:
RuntimeError: function signature mismatch
    at wa-sqlite.wasm:wasm-function[517]:0x30368
    at wa-sqlite.wasm:wasm-function[830]:0x4550b
    at wa-sqlite.wasm:wasm-function[1217]:0x65d91
    at wa-sqlite.wasm:wasm-function[1649]:0x7b6b1
    at Object.changeset_apply (wa-sqlite/src/sqlite-api.js:1176:22)
    at Object.apply (sqlite-wasm/src/make-sqlite-db.ts:189:21)
    at SqliteDbWrapper.rollback (livestore/src/SqliteDbWrapper.ts:94:23)
    at rollback (livestore/src/store/store.ts:188:52)
    at ClientSessionSyncProcessor.ts:183:17
    at client-session-sync-processor:pull (ClientSessionSyncProcessor.ts:235:14)

Then LiveStore shuts down:
[@livestore/adapter-web:client-session] client-session shutdown
```

### Additional Context

Happy to try resolving this bug but would need some guidance in best path forward as it touches critical code-paths and potentially C code which I'm less familiar with. 

### Checklist

acusti:
## Summary

Fixes #998

`sqlite3.changeset_apply` passes a conflict handler callback to the C function `sqlite3changeset_apply`. The current code passes a plain JS function, but `cwrap` declares all parameters as `'number'`, so the function is silently coerced to `0` (null pointer). This causes a `RuntimeError: function signature mismatch` whenever SQLite invokes the conflict callback during rebase rollback.

### Changes

**`packages/@livestore/wa-sqlite/src/sqlite-api.js`**

- Use `Module.addFunction()` to register the conflict handler as a proper WASM function pointer
- The pointer is created **lazily** on the first `changeset_apply` call and cached for reuse, so environments that disallow dynamic WebAssembly compilation (e.g. Cloudflare Workers) don't crash during `Factory()` init if `changeset_apply` is never called

**`packages/@livestore/wa-sqlite/Makefile`**

- Add `-sALLOW_TABLE_GROWTH=1` to `EMFLAGS_COMMON` so the WASM function table can accommodate the `addFunction` call (the current binaries ship with `table max == table min`, leaving zero room for runtime function registration)

### Notes

- The WASM binaries will need to be rebuilt with the updated Makefile for the fix to work end-to-end. Without `ALLOW_TABLE_GROWTH`, `addFunction` fails because the function table can't grow.
- The lazy initialization is important: `addFunction` calls `convertJsFunctionToWasm`, which dynamically compiles a small WASM module. This is fine in browsers and Node.js, but Cloudflare Workers disallow dynamic `WebAssembly.Module()` construction. By deferring to first use, environments that import but never call `changeset_apply` (e.g. Durable Objects using native SQLite) aren't affected.

## Test plan

- [x] Verified fix resolves the `RuntimeError: function signature mismatch` crash in browser
- [x] Verified fix works in Cloudflare Durable Objects (no `CompileError: Wasm code generation disallowed by embedder`)
- [x] Verified all existing tests pass with the patched source + WASM table growth enabled

## Responsible Disclosure

The code changes and description in this PR were created by Claude Code Opus 4.6. I reviewed the changes but I don’t have any experience with WASM, so I’m way out of my depth here. The first version of the fix for this created the onConflictPtr (`const onConflictPtr = Module.addFunction(...)`) directly in the IIFE (above the returned function), like:
```js
    const onConflictPtr = Module.addFunction(
      (_pCtx, _eConflict, _pIter) => SQLITE_CHANGESET_REPLACE,
      'iiii'
    );

    return function(db, changesetData, options) {
```
but that triggered errors in my DO client: `Error: (FiberFailure) CompileError: WebAssembly.Module(): Wasm code generation disallowed by embedder`

Hence the lazy initialization approach taken here, which keeps the DO from crashing as long as it doesn’t hit the `changeset_apply` codepath.


acusti:
should i rebase this PR and resolve conflicts?

IGassmann:
Closing in favor of #1173, which fixes the same issue using the established C adapter relay pattern (`libsession.c`/`libsession.js`) instead of `Module.addFunction`. This avoids the need for `-sALLOW_TABLE_GROWTH=1` and dynamic WASM compilation, stays consistent with how every other callback in wa-sqlite (`commit_hook`, `update_hook`, `create_function`) is wired, exposes both `xFilter` and `xConflict` as explicit parameters matching the `sqlite3changeset_apply` C API, and supports custom conflict handlers.

Thank you @acusti for the investigation and initial fix!

PR #1029, opened 2026-02-15
fix: sync deadlock during paginated pulls in `LeaderSyncProcessor`

slashv:
## Preface

Full disclosure, Codex 5.3 was used to both identify and solve this bug. I was experiencing a sync dead-lock in my LiveStore app when doing heavy bursts of events at startup. After implementing this and running my app against LiveStore with this fix it solved the problem.

I have a conceptual understanding of why this occurred as well as the solution but I have not written any code by hand nor have I examined every line in detail. 

## Problem

What problem does this pull request address?

- `pushPullMutex` could remain held on some failure/exit paths.
- Paginated pull could re-acquire before release, causing stalls/deadlock symptoms.

## Solution

How does this change resolve the problem? Mention trade-offs and alternatives considered.

- Refactor local push critical section to scoped semaphore usage (`withPermits(1)`).
- In pull path, track lock ownership across pages and guarantee release on:
  - terminal page (`NoMore`)
  - chunk failure
  - interruption/stream teardown (`ensuring`)
- Add regression coverage for paginated non-live pull + post-pull local push progress.

## Validation

What did you do to verify the change?

- `direnv exec . vitest run tests/package-common/src/leader-thread/LeaderSyncProcessor.test.ts`
- `direnv exec . mono ts`
- `direnv exec . mono lint --fix`

### Demo (optional)

- Before: pending changes could stall indefinitely.
- After: paginated pull completes and local push resumes; regression test passes.


schickling-assistant:
Thanks for tackling this - the mutex lifecycle changes look solid and the paginated non-live regression test is very helpful.

One coverage gap I still see: could we add tests for lock release on non-happy paths during paginated pull?
- pull chunk failure while the mutex is already held (mid-pagination)
- interruption/cancellation of the pull stream while pagination is in progress

Those seem like the most subtle regressions for this area and would give stronger confidence that  +  always prevent stalls.

Also, could you please rebase this branch onto latest  and fix the current lint issue before merge?

Thanks again for driving this fix.

schickling-assistant:
Small follow-up/correction (my previous comment had formatting glitches from CLI escaping):

The two symbols I referenced are releasePullMutexIfHeld and Effect.ensuring.

Request remains the same:
- please add coverage for mid-pagination pull failure and interruption/cancellation while the mutex is held
- please rebase on latest dev and fix the lint issue

Thanks again for the fix and for following up on this.

slashv:
Re-built this fix on top of #1099, so it now keeps that local-push mutex cleanup/rename and narrows this PR to the remaining pull-side fix.

This update also addresses the review feedback by adding coverage for the two paginated pull failure paths that were called out:

- mid-pagination pull failure releases the mutex
- mid-pagination pull interruption/cancellation releases the mutex

I also rebased onto latest dev and re-ran tests/package-common/src/leader-thread/LeaderSyncProcessor.test.ts successfully.

IGassmann (approved):

PR #1070, opened 2026-02-28
fix(live-queries): detect hermes bytecode strings with arguments

### LINKED ISSUE #1000: [Bug]: ReactNative / Expo queries has same hash key when function passed to `queryDb`
### Bug Description

 `queryInput.toString()` in these lines:
https://github.com/livestorejs/livestore/blob/ff8425a5f224227d5493e753b90717a3a8d46829/packages/%40livestore/livestore/src/live-queries/db-query.ts#L167-L169

Always produces `function (a0) { [bytecode] }` on react native/expo when using arrow function causing to use same cached result for all queries that use `queryDb(get => {...})`

Workaround is to pass unique values `deps` which is used when creating hash, to use named functions like `queryDb(function uniqueName(get)  {...})` or set `'show source'` in function body like:
```tsx
export const todos$ = queryDb(
  get => {
    'show source'; // hermes directive to include function body (https://github.com/facebook/hermes/issues/114)

  })
```

### Minimal Reproduction

I'm willing to create fresh expo project if asked, this could be marked as a documentation issue since we it might not be easy fix unless we add a babel plugin or something.

### LiveStore Version

0.4.0-dev.22

_No response_

### Environment

```markdown
N/A
```

### Expected Behavior

Use of `queryDb(get => {...})` should not result in same cached results for all queries

### Actual Behavior

Use of `queryDb(get => {...})` results in same data in all queries.

### Error Logs/Screenshots

```shell

```

### Additional Context

_No response_

### Checklist

A386official:
Fixes #1000

Update isValidFunctionString in function-string.ts to check for the `[bytecode]` substring (or a regex like /function.*\[bytecode\]/) rather than an exact string match, so the existing guard at db-query.ts:110 properly throws when running under Hermes with function-input queries that lack deps

Added a test case to cover the fix.


A386official:
Closing — this PR inadvertently included upstream changes resulting in an oversized diff. Apologies for the noise.

PR #1077, opened 2026-03-03
feat(adapter-cloudflare): expose initialSyncOptions in createStoreDo

bohdanbirdie:
The initial sync timeout in the Cloudflare DO adapter was hardcoded to 500ms. For DOs with large eventlogs, this isn't enough time to complete the initial sync before the timeout kicks in.

This adds initialSyncOptions to CreateStoreDoOptions so the timeout can be configured per-deployment, or switched to skip mode entirely.


bohdanbirdie:
@IGassmann can this minor change go through?
I'll rebase once approved

IGassmann (approved):

Issue #829, opened 2025-11-17
Separate VFS contract from LiveStore-specific utilities


IGassmann:
## Problem

There are two related architectural issues with the current AccessHandlePoolVFS implementation:

1. **Duplicated utilities**: `opfs-sah-pool.ts` contains utilities like `decodeAccessHandlePoolFilename` that re-implement concerns already handled by `AccessHandlePoolVFS`, creating duplication and breaking encapsulation.

2. **Mixed concerns**: `AccessHandlePoolVFS` mixes SQLite VFS contract methods (like `jOpen`, `jRead`, `jWrite`) with application-specific utilities (like `getOpfsFileName()`, `resetAccessHandle()`, `readFilePayload()`, `getTrackedFilePaths()`), making the code harder to understand and maintain.

## Solution

Following the pattern from sqlite-wasm's `SAHPoolUtil` (see https://github.com/sqlite/sqlite-wasm/blob/main/index.d.ts#L1276), separate these concerns:

**Keep in VFS class (SQLite contract):**
- VFS contract methods: `jOpen`, `jClose`, `jRead`, `jWrite`, `jTruncate`, `jSync`, `jFileSize`, `jAccess`, `jDelete`, etc.
- Core VFS lifecycle: `isReady()`, `close()`

**Move to utility class (similar to SAHPoolUtil):**
- File header operations: `decodeAccessHandlePoolFilename()` from opfs-sah-pool.ts
- File name mapping: `getOpfsFileName()`
- Application-specific cleanup: `resetAccessHandle()`
- Payload extraction: `readFilePayload()` (for migration)
- Introspection: `getTrackedFilePaths()`
- Pool management: `addCapacity()`, `removeCapacity()`

## Benefits

- Eliminates code duplication between opfs-sah-pool.ts and AccessHandlePoolVFS
- Clearer separation between SQLite VFS protocol and LiveStore-specific utilities
- Easier to understand which methods are part of the VFS contract
- Better testability with focused responsibilities
- Aligns with sqlite-wasm's established pattern

## Files Involved

- `packages/@livestore/sqlite-wasm/src/browser/opfs/AccessHandlePoolVFS.ts` — VFS implementation
- `packages/@livestore/sqlite-wasm/src/browser/opfs/opfs-sah-pool.ts` — utilities to consolidate

## Related

Follow-up from #653

See also: https://github.com/sqlite/sqlite-wasm/blob/main/index.d.ts#L1276

Issue #714, opened 2025-10-01
Resume Cloudflare DO pulls after mid-stream socket close


schickling:
## Summary
Durable Objects close the Cloudflare WebSocket while a historical pull is still streaming, leaving ~3.5k events unserved. The client keeps waiting on the `pull` stream and our Vitest suite times out after 60 s.

## Timeline (high-level)
```
Client pull(cursor=None)
        │
        ▼
DO streams chunk #175 (remaining ≈3.5k)
        │
        ├── Cloudflare runtime cleanly closes socket (code 1000)
        ▼
DO pull fiber exits via interrupt (“All fibers interrupted without errors”)
        │
        └── Client waits forever → test times out after 60 s
```

## Reproduction
```
CI=1 direnv exec . vitest run tests/sync-provider/src/sync-provider.test.ts \
  --testNamePattern "'Cloudflare WebSocket (DO)'"
```
This consistently surfaces the stall after ~7 k events during the “large batch pagination” scenario.

## Evidence
All logs captured on 2025-10-01 using wrangler dev:

```
# DO streaming chunk #174 with partial page
TMP do pull res {"backendId":"pIZBOyisy9KQ0RORWnU-v","chunkIndex":174,"batchSize":56,"firstSeq":26344,"lastSeq":26399,"pageInfo":{"_tag":"MoreKnown","remaining":3600}}
TMP do storage page {"cursor":26399,"limit":256,"rawCount":256}
TMP do storage next {"nextCursor":26655,"nextLimit":256}

# Next chunk emitted; runtime closes immediately afterwards
TMP do pull res {"backendId":"pIZBOyisy9KQ0RORWnU-v","chunkIndex":175,"batchSize":100,"firstSeq":26400,"lastSeq":26499,"pageInfo":{"_tag":"MoreKnown","remaining":3500}}
TMP do ws close {"storeId":"test-store-Cloudflare WebSocket (DO)-large-batch-test-cFH-FM0hsUmqFTLKe5I6Y","pullRequestIds":["180"],"pullRequestCount":1,"code":1000,"wasClean":true}
TMP do pull stream exit {"storeId":"test-store-Cloudflare WebSocket (DO)-large-batch-test-cFH-FM0hsUmqFTLKe5I6Y","backendId":"pIZBOyisy9KQ0RORWnU-v","exit":"interrupt","chunks":76,"events":6500,"cause":"All fibers interrupted without errors."}
```

[...]

Issue #562, opened 2025-08-16
Async SQLite API for Large Dataset Support Without Full In-Memory Requirement


schickling:
# Async SQLite API for Large Dataset Support Without Full In-Memory Requirement

## Overview

Currently, LiveStore loads the entire SQLite database into memory for each client session (browser tab, worker, etc.) to provide synchronous query execution and instant reactivity. While this approach delivers exceptional performance for typical local-first applications, it creates limitations when dealing with larger datasets or resource-constrained environments.

This issue proposes implementing an **async SQLite API** that allows LiveStore to work with databases that exceed practical memory limits while maintaining the core benefits of the framework.

### Current Limitations

1. **Memory Constraints**: Each client session requires loading the full database into memory
   - Browser tabs have memory limits (typically 2-4GB)
   - Mobile devices have even stricter constraints
   - Multiple tabs multiply memory usage

2. **Large Dataset Challenges**:
   - Applications with extensive historical data
   - Media-rich applications with embedded blobs
   - Analytics/reporting use cases with large fact tables
   - Multi-tenant applications with consolidated data

3. **Resource Efficiency**:
   - Unnecessary memory usage for rarely accessed data
   - Cold data taking up valuable RAM
   - Background tabs maintaining full database copies

## Commercial Sponsorship Required

**⚠️ This is a major architectural enhancement that requires significant development resources.** 

**We are looking for commercial project sponsors to fund this development.** If your company would benefit from async SQLite support and is interested in sponsoring this feature, please get in touch via email or Discord.

### Benchmarks Needed
- Memory usage reduction for large datasets
- Query latency for async vs sync operations  
- Battery impact on mobile devices
- Network overhead for partial sync

### Optimization Strategies
- **Query result caching**: LRU cache for frequent queries
- **Prefetching**: Predictive loading based on access patterns
- **Indexes on disk**: Maintain indexes for async tables
- **Columnar storage**: For analytical workloads

## Success Criteria

- [ ] Support databases >1GB without loading fully into memory
- [ ] Async queries complete in <100ms (?) for indexed operations
- [ ] Seamless sync protocol integration

## Related Issues

- Supersedes #188 - "Store and retrieve data without the memory sqlite database"
- Related to #266 - "Cloudflare Worker Adapter" (D1 async operations)
- Related to #356 - "SQLite compatibility test suite" (async operations testing)
- Related to #468 - "Pluggable SQLite drivers" (async driver support)

## Open Questions

[...]

Issue #409, opened 2025-07-03
Crash in materializer while syncing new events advances `backendHead` and causes later boot failure


ldirer:
I noticed that LiveStore advances the backendHead in `__livestore_sync_status` before materialization. 

This can lead to inconsistencies if a crash occurs when materializing: the eventlog and materialized state are not updated, but the `backendHead` is. 
Once in this state the node adapter will crash on startup (even after removing the faulty materialization code) with:
```
During boot the backend head (6) should never be greater than the local head (5)
[18:33:50.743 window] ERROR (#17): LiveStore.UnexpectedError: { "cause": Error: This should never happen: During boot the backend head (6) should never be greater than the local head (5), "note": undefined, "payload": undefined }
```

I could only resolve it by manually editing the `__livestore_sync_status`.  

## Reproduction Steps 

So this is a little tricky, we need faulty code to run materialization on newly synced events, so there needs to be another writer.  
But if the other writer has faulty code, it won't be able to write.  

I pushed a [branch](https://github.com/ldirer/livestore/tree/repro-materializer-crash-backend-head-inconsistency) to help with the repro, but it might just be as quick to make the changes locally. It has invalid code to begin with, to confirm  the error persists after fixing the code you can `git revert HEAD` after step 4 and try again. 

I reproduced this by using both `examples/standalone/web-todomvc-sync-cf`  and `examples/standalone/node-todomvc-sync-cf`.  
We use the first one as the "good writer" and the second one as the "faulty reader". 

0. Modify `examples/standalone/node-todomvc-sync-cf/src/livestore/schema.ts:42`  to return "noop" from the `TodoCreated` materializer (this is done in my repro branch, a one-line change).

1. Start the good writer: `cd examples/standalone/web-todomvc-sync-cf; bun dev` 
2. Go to [http://localhost:60001/?storeId=a](http://localhost:60001/?storeId=a), create a todo.
3. Start the faulty reader: `cd examples/standalone/node-todomvc-sync-cf; STORE_ID=a bun start`
This should crash: this is expected and fine.
4. Run `STORE_ID=a bun start` again.
It should crash with "this should never happen ...". 
**repro achieved** 🎉 

Note the error at step 4 is a different error from step 3, and fixing the code does not fix it.
This repro isn't very "minimal", not sure how to do much better.

## Investigation

This doesn’t happen for all errors:

- Missing materializer: crash, but backend head is not advanced ✅
[...]

PR #1089, opened 2026-03-10
perf(adapter-cloudflare): reduce billable rows_written by 95%

bohdanbirdie:
## Problem

The `@livestore/adapter-cloudflare` stores both `dbEventlog` and `dbState` using wasm SQLite with `CloudflareSqlVFS`. This VFS persists SQLite pages as rows in a `vfs_blocks` table on DO SQLite ([SQLite in Durable Objects](https://developers.cloudflare.com/durable-objects/api/sqlite-storage-api/)), splitting data into 64 KiB blocks. Every logical SQLite write triggers multiple `INSERT OR REPLACE` operations on `vfs_blocks` — one per affected block.

This causes **~238x write amplification**: a single event commit that should cost 1 `rows_written` costs ~238 instead. Cloudflare bills DO SQLite usage by `rows_written` (free tier: 100k/day). With the current VFS approach, even moderate usage quickly exhausts the free tier. Cold starts cost ~4,003 writes due to full eventlog rematerialization through the VFS.

## Solution

Fix the existing VFS instead of replacing it. The original VFS architecture is sound (disk-backed, low memory, no snapshot needed, rebase works via wasm session extension). The 238x write amplification came from two root causes:

### 1. Block/page size mismatch → read-merge-write amplification

`BLOCK_SIZE` was 64KB (inherited from `CloudflareWorkerVFS` which used `storage.put()` with a 128KB value limit). `dbState` uses 8KB pages (`PRAGMA page_size=8192`). Every page write triggered a read-merge-write of the full 64KB block via `mergePartialBlock`.

**Fix:** Change `BLOCK_SIZE` to `8 * 1024` in `CloudflareSqlVFS.ts`. One page = one block, no merging. The `storage.sql` API has no size limit — the 64KB block size was unnecessary.

### 2. Journal files written through VFS

Every wasm SQLite transaction created, wrote, and deleted a journal file through the VFS to DO SQLite. Journal writes are non-page-aligned (4-byte page number + page data + 4-byte checksum), causing additional `mergePartialBlock` overhead.

**Fix:** `PRAGMA journal_mode=MEMORY` on `dbState` keeps the rollback journal in wasm heap memory. In-process transaction ROLLBACK still works. Crash recovery via journal was never functional in this VFS — no replay logic exists, and VFS block writes aren't wrapped in `storage.transactionSync()`.

### 3. Redundant index

`idx_vfs_blocks_range ON vfs_blocks(file_path, block_id)` duplicated the composite PRIMARY KEY, doubling `rowsWritten` per block write.

**Fix:** Removed the index creation. The PRIMARY KEY already serves as the clustered index.

### Additional changes

- **`dbEventlog`** moved to DO SQLite via `makeDoSqliteDb` (`_tag: 'file'`). 1 event = 1 `rows_written`, bypassing VFS entirely.
- **`PERSISTENT_FILE_TYPES`** changed to only `SQLITE_OPEN_MAIN_DB`. Journal/WAL files are transient — persisting them added unnecessary `rowsWritten`.
- **Transaction no-op** in `make-sqlite-db.ts`: `isTransactionControlStatement()` skips `BEGIN`/`COMMIT`/`ROLLBACK`/`SAVEPOINT`/`RELEASE` on DO SQLite (CF rejects SQL-level transaction control). Safe due to single-threaded DO model.

All changes are confined to `@livestore/adapter-cloudflare` and `@livestore/sqlite-wasm` — no modifications to `@livestore/common` or other packages.

### Trade-offs

- `dbEventlog.export()` returns empty — only used by disabled devtools, not a concern for production DOs.
- `journal_mode=MEMORY` eliminates crash recovery via journal, but this was never functional in `CloudflareSqlVFS` (no hot journal detection or replay logic exists).
- 12 writes/todo vs the snapshot approach's 4.2 — trades ~8 extra writes/event for no CPU scaling bottleneck, no memory pressure, and no snapshot mechanism.

## Approach evolution

The initial commits used an in-memory `dbState` with snapshot persistence (`dbState.export()` after every COMMIT). This achieved 4.2 writes/todo but hit an O(total DB size) CPU bottleneck — `dbState.export()` serializes the entire in-memory DB, causing request timeouts at ~10k state rows on CF. The in-memory DB also lives entirely in the DO's 128MB RAM limit.

The current approach fixes the VFS directly, preserving its architectural advantages (disk-backed, low memory, no snapshot, rebase works) while eliminating the write amplification.

| Metric | Original VFS | Snapshot approach | Fixed VFS (current) |
|---|---|---|---|
| Writes/todo (individual) | ~238 | ~6 | **12** |
| Writes/todo (bulk) | ~238 | ~4.2 | **5.5** |
| Cold start writes | ~4,003 | ~8 | **0** |
| CPU per COMMIT | O(1) | O(total DB size) | **O(dirty pages) ≈ O(1)** |
| Memory | Low (page cache) | All state in RAM | **Low (page cache)** |
| Scaling limit | None | ~10k state rows | **None** |

## Validation

Integration tests run on a real wrangler dev server with actual DOs and `storage.sql`:

| Scenario | Before (VFS) | After (fixed VFS) | Reduction |
|---|---|---|---|
| Steady-state per todo | 238 | **12** | 95.0% |
| Cold start writes (5 todos) | ~4,003 | **0** | 100% |
| Cold start reads | — | **21** | — |
| Post-restart per todo | 238 | **12** | 95.0% |
| Data survives shutdown cycles | ✓ | ✓ | — |

6 test cases covering:
1. Durable Object state persists when `resetPersistence` is not requested
2. Durable Object persistence clears when `resetPersistence` is true
3. Steady-state writes (asserts < 20 writes/todo, measured: 12)
4. Cold start reopens persisted VFS state with zero writes (measured: 0 writes, 21 reads)
5. Data survives multiple shutdown cycles
6. Post-restart steady-state writes remain low (measured: 12 writes/todo)

### Production benchmark (deployed to CF, ~1,101 events)

| Metric | Local | Production |
|---|---|---|
| Boot cost | 117 writes | 117 writes |
| Steady-state (bulk) | 5.5/todo | 5.1–6.2/todo |
| Cold start | 14 writes | 14 writes |
| CF dashboard rows_written | — | **6.05k total** |
| Exceeded memory limits | — | **0** |
| Exceeded CPU limits | — | **0** |

At 12 writes/todo, the CF free tier (100k rows_written/day) supports ~8,300 events/day — up from ~420 with the original VFS.

## Known limitation

The sync cursor (`syncHead`) lags behind after burst writes — the background pull task doesn't get enough CPU time between HTTP requests because CF evicts idle DOs. This is a **pre-existing issue** in the CF adapter (not introduced by this PR) and affects multi-client sync convergence. Will file a separate issue.

## Related issues

- #355 — DO SQLite binding for Node.js adapter (same concept, different target)
- #137 — Exchangeable database drivers
- #829 — Separate VFS contract from LiveStore-specific utilities
- #142 — Replace AccessHandlePoolVFS with a more purpose-built SqliteVFS
- #356 — Comprehensive SQLite compatibility test suite (this PR adds CF adapter coverage)
- #562 — Async SQLite API for large datasets
- #714 — DO 30s CPU eviction (fewer writes = faster processing, reduced risk)
- #409 — Backend head > local head crash
- #694 — Guard sqlite imports against WAL snapshots

<details>
<summary>Previous PR description (snapshot approach — superseded)</summary>

## Problem

The `@livestore/adapter-cloudflare` stores both `dbEventlog` and `dbState` using wasm SQLite with `CloudflareSqlVFS`. This VFS persists SQLite pages as rows in a `vfs_blocks` table on DO SQLite, splitting data into 64 KiB blocks. Every logical SQLite write triggers multiple `INSERT OR REPLACE` operations on `vfs_blocks` — one per affected block.

This causes **~238x write amplification**: a single event commit that should cost 1 `rows_written` costs ~238 instead. Cloudflare bills DO SQLite usage by `rows_written` (free tier: 100k/day). With the current VFS approach, even moderate usage quickly exhausts the free tier. Cold starts cost ~4,003 writes due to full eventlog rematerialization through the VFS.

## Solution

Replace VFS-backed storage with DO SQLite for `dbEventlog` and in-memory wasm SQLite (with snapshot persistence) for `dbState`:

- **`dbEventlog`**: uses `makeDoSqliteDb` with `_tag: 'file'` — writes go directly to DO SQLite tables, bypassing VFS block splitting. 1 event = 1 `rows_written`.
- **`dbState`**: runs in-memory wasm SQLite (zero `rows_written` during materialization). After each materialization batch COMMIT, the in-memory DB is exported as a blob and persisted to a `_state_snapshot` table (1 row written).
- **Transaction no-op**: `isTransactionControlStatement()` in `make-sqlite-db.ts` skips `BEGIN`/`COMMIT`/`ROLLBACK`/`SAVEPOINT`/`RELEASE` since CF DO SQLite rejects SQL-level transaction control and requires `storage.transactionSync()` instead. Safe because the single-threaded DO model guarantees no concurrent access.
- **Snapshot restore**: On cold start, loads snapshot blob from `_state_snapshot`, imports into in-memory DB, and truncates local eventlog cache if snapshot head is behind — sync processor re-pulls the delta from `SyncBackendDO`.

Changes are confined to `@livestore/adapter-cloudflare` — no modifications to `@livestore/common` or other packages.

### Trade-offs

- `dbEventlog.export()` returns empty — only used by disabled devtools, not a concern for production DOs.
- Snapshot save runs on every materialization batch COMMIT via `dbState.execute` interception — no post-materialization hook exists in livestore, so this is the only integration point. For typical DO workloads (small state DBs), export + write is sub-millisecond.
- Schema changes (new `stateDbFileName` via hash) naturally invalidate snapshots → full in-memory rematerialization.

## Validation

The first commit (`4a83b99a6`) adds the integration test infrastructure and runs against the **existing VFS-backed adapter** to establish the baseline (~238 writes/todo). The second commit (`e5a444fca`) applies the optimization and updates assertions to verify the new low write counts. This before/after approach validates the improvement on the same test harness.

Integration tests run on a real wrangler dev server with actual DOs and `storage.sql`:

| Scenario | Before (VFS) | After (DO SQLite + snapshot) | Reduction |
|---|---|---|---|
| Boot + first todo | 1,113 | **18** | 98.4% |
| Steady-state per todo | 238 | **6** | 97.5% |
| 10 todos total | 2,380 | **60** | 97.5% |
| Cold start (snapshot restore) | ~4,003 | **8** | 99.8% |

4 test cases covering:
1. Steady-state writes (asserts < 20 writes/todo, measured: 6)
2. Snapshot restore on cold start (asserts < 50 writes, measured: 8)
3. Data survives multiple shutdown cycles
4. Post-restart steady-state writes remain low

TypeScript strict check and lint pass.

## Related issues

- #355 — DO SQLite binding for Node.js adapter (same concept, different target)
- #137 — Exchangeable database drivers
- #829 — Separate VFS contract from LiveStore-specific utilities (this change bypasses VFS entirely)
- #142 — Replace AccessHandlePoolVFS with a more purpose-built SqliteVFS
- #356 — Comprehensive SQLite compatibility test suite (this PR adds CF adapter coverage)
- #562 — Async SQLite API for large datasets
- #714 — DO 30s CPU eviction (fewer writes = faster processing, reduced risk)
- #409 — Backend head > local head crash (our eventlog truncation resets heads correctly)
- #694 — Guard sqlite imports against WAL snapshots

</details>


tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
```diff
@@ -0,0 +1,21 @@
+name = "adapter-cloudflare-writes-test"
+main = "./worker.ts"
+compatibility_date = "2024-05-12"
```
IGassmann: Why did you opt for such an old compatibility date?

tests/integration/src/tests/adapter-cloudflare-writes/adapter-cloudflare-writes.test.ts
IGassmann, line 1: Why did you opt for creating this new test file and all its related files instead of writing the new test in the existing `tests/integration/src/tests/adapter-cloudflare/adapter-cloudflare.test.ts` file?

tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
```diff
@@ -0,0 +1,21 @@
...
+compatibility_flags = ["enable_request_signal"]
```
IGassmann: Add comment describing why this compatibility flag is enabled

tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
```diff
@@ -0,0 +1,21 @@
...
+class_name = "WriteTrackingStoreDo"
+
+[[migrations]]
+tag = "v1"
+new_sqlite_classes = ["SyncBackendDO", "WriteTrackingStoreDo"]
+
+[[d1_databases]]
+binding = "DB"
+database_name = "livestore-writes-test"
+database_id = "unused"
```
IGassmann: Why does this D1 database exists?

packages/@livestore/adapter-cloudflare/src/make-adapter.ts
```diff
@@ -17,6 +18,7 @@ import {
 } from '@livestore/common/leader-thread'
 import { LiveStoreEvent } from '@livestore/livestore'
 import { sqliteDbFactory } from '@livestore/sqlite-wasm/cf'
+import { makeSqliteDb as makeNativeSqliteDb } from './make-sqlite-db.js'
```
IGassmann:
```suggestion
import { makeSqliteDb as makeNativeSqliteDb } from './make-sqlite-db.ts'
```

bohdanbirdie:
## Production Benchmark: Memory & Limits Validation (CF deployed, 2026-03-15)

Deployed a bench worker to Cloudflare to validate whether the in-memory `dbState` approach hits the 128MB DO memory limit under load.

### Results (real CF infrastructure)

| Total todos in store | writes/todo | Exceeded memory | Exceeded CPU |
|---------------------|-------------|-----------------|--------------|
| 10 | 4.6 | 0 | 0 |
| 100 | 4.2 | 0 | 0 |
| 1,000 | 4.2 | 0 | 0 |
| 5,000 | 4.2 | 0 | 0 |
| 10,000 | 4.2 | 0 | 0 |

Cold start with 10k todos: **6 rows_written**, all data intact.

A second run pushing to 20k+ showed no memory errors either, but requests began timing out due to snapshot export CPU cost — not a memory issue.

### Known scaling limitation: snapshot export cost

`dbState.export()` serializes the **entire** in-memory SQLite DB on every materialization COMMIT. As the store grows, each export becomes more expensive — regardless of whether it's a bulk insert or a single event. At 10k+ rows, this adds noticeable latency to every write. At 20k+, it causes request timeouts in production.

This is not a correctness issue — data is safe and the eventlog is always durable. But it's a latency concern for stores that grow large. Possible mitigations:
- **Throttle snapshot saves** (every N commits or time-based) — reduces export frequency at the cost of more re-materialization on crash
- **Incremental persistence** (dirty page tracking) — only persist changed pages, constant CPU regardless of DB size
- **Save on DO eviction only** (`blockConcurrencyWhile` + alarm) — zero snapshot cost during normal operation

### Conclusion

The in-memory `dbState` with snapshot persistence does **not** hit the 128MB memory limit at 10k+ todos. The practical scaling bottleneck is CPU time from full snapshot export on every COMMIT. For typical DO workloads (< 10k events per store), this works well. For larger stores, snapshot strategy optimization is needed as a follow-up.

### Bench app

The benchmark is committed at `tests/cf-bench/` — it's a standalone project with its own `pnpm-workspace.yaml` and lockfile, independent from the monorepo. See its README for usage.

tests/integration/src/tests/adapter-cloudflare-writes/adapter-cloudflare-writes.test.ts
bohdanbirdie: The test needs a custom `WriteTrackingStoreDo` that wraps `storage.sql` with a Proxy to accumulate `cursor.rowsWritten` after every `exec()` call. This requires different DO class bindings in `wrangler.toml` and a different worker entry point. The existing `TestStoreDo` tracks VFS persistence counts (files/blocks), not `rowsWritten` - they measure different things with different infrastructure.

tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
bohdanbirdie: Same pattern as the existing `adapter-cloudflare` test (`tests/integration/src/tests/adapter-cloudflare/fixtures/worker.ts:41`). The DB binding is declared in the Env type but never used in either worker. Happy to remove it from both if requested

tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
bohdanbirdie: Matches the existing adapter-cloudflare integration test (`tests/integration/src/tests/adapter-cloudflare/fixtures/wrangler.toml:3`).
Should I update it in both?

tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
IGassmann: Let's remove it then if it's not used

tests/integration/src/tests/adapter-cloudflare-writes/fixtures/wrangler.toml
IGassmann: Yes, please :)

bohdanbirdie:
Updated the PR title and description to reflect a changed approach: **fixing the existing VFS** instead of replacing it with snapshots.

The snapshot approach (initial commits) achieved 4.2 writes/todo but hit an O(total DB size) CPU bottleneck from `dbState.export()` after every materialization COMMIT — requests timed out at ~10k state rows on CF. The fixed VFS approach (latest commit) achieves 14 writes/todo with no scaling bottleneck, no memory pressure, and no snapshot mechanism.

Production benchmark on a deployed CF worker confirms all numbers match local results. See updated description for full details and comparison table.

The previous description is preserved in a collapsed section at the bottom.

bohdanbirdie:
**Sync investigation findings (stress testing with 1000+ events)**

While stress-testing the VFS fix using the `cf-bench` app (updated with sync-status tracking and better error reporting in this PR), I found two pre-existing sync issues that cause data loss on cold start. Neither is introduced by this PR, but worth flagging:

**1. `pushPullMutex` deadlock after store shutdown + recreation**

After `store.shutdownPromise()` + new `createStoreDoPromise()` within the same DO, the pull stream doesn't release the mutex on interruption. `backgroundApplyLocalPushes` blocks forever. Events commit in-memory but never reach `dbEventlog`/`dbState`. Data lost on DO eviction.

I validated that #1029 resolves this.

**2. `livePull` defaults to `false`, `syncHead` never advances**

With `livePull = false` (current default in `createStoreDo`), the pull request is sent without `rpcContext`, so the sync backend never registers the client as a subscriber. Pushed events succeed but are never echoed back, `updateBackendHead()` never fires, `syncHead` stays at 0. On cold start with stale `syncHead`, all events are treated as pending, leading to destructive rebase that wipes pre-shutdown state.

Setting `livePull = true` fixes this. I validated locally with 1500+ events surviving shutdown/restart with zero data loss. The `false` default might be confusing since it silently breaks sync persistence; maybe it should be enabled by default?

bohdanbirdie:
Moved `cf-bench` to a separate PR: https://github.com/livestorejs/livestore/pull/1106

IGassmann (approved):
Great work!

PR #1099, opened 2026-03-16
refactor: scope `localPushBackendPullMutex` with `withPermits`

IGassmann:
## Problem

The `pushPullMutex` in `backgroundApplyLocalPushes` used manual `take(1)` / `release(1)` calls across three separate exit paths. This was error-prone — the `rebase` case (`dieDebugger`) didn't release the mutex at all, and adding new code paths required remembering to release manually.

The name `pushPullMutex` was also vague about what it actually guards.

## Solution

- Replace the manual acquire/release pairs with a single scoped `withPermits(1)` via `.pipe()`. Early exits become `return` from the inner `Effect.gen`, and `withPermits` guarantees release on any exit (success, failure, die, interrupt).
- Rename `pushPullMutex` → `localPushBackendPullMutex` to clarify the two sides it serializes.

## Validation

Structural refactor — existing sync tests cover push/pull mutual exclusion behavior. No new tests needed.

## Related issues

None.

### DIFF
--- a/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
+++ b/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
@@ -184,7 +184,7 @@ export const makeLeaderSyncProcessor = ({
 
     const localPushesQueue = yield* BucketQueue.make<LocalPushQueueItem>()
     // Ensures mutual exclusion between local push and backend pull processing.
-    const pushPullMutex = yield* Effect.makeSemaphore(1)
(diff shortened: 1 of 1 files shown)


slashv:
#1029 adds the pull-side half of the mutex fix that #1099 doesn’t cover, ensuring paginated backend pulls always release the shared lock on completion, failure, or interruption and proving it with regression tests. 

Pulled in your exact changes, so #1029 is now built directly on top of #1099.

IGassmann:
Superseded by https://github.com/livestorejs/livestore/pull/1029

PR #1106, opened 2026-03-19
Add cf-bench: standalone CF adapter benchmark app

bohdanbirdie:
Extracted from #1089 to keep that PR focused on the VFS write amplification fix.

- Adds a standalone deployable Cloudflare Worker (`tests/cf-bench`) for benchmarking the livestore DO adapter's `rows_written` and cold start performance on real CF infrastructure
- Includes `package.json.genie.ts` with proper `catalog.compose()` for genie coverage
- Adds `tests/cf-bench/**` to oxlint ignore patterns (standalone workspace, types unavailable in CI)

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -24,6 +24,7 @@
     "**/patches/**",
     "**/.cache/**",
     "**/.turbo/**",
+    "tests/cf-bench/**",
     "tests/integration/node_modules/**",
     "docs/src/plugins/**"
   ],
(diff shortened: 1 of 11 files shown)


bohdanbirdie:
@IGassmann do you think this is still valuable after the DO adapter fix was merged?

IGassmann:
I think the benchmark itself was valuable for validating the fix, but I don't think it fits well with our existing testing stack as a standalone deployable worker with its own pnpm-lock.yaml and workspace. If we want to keep regression coverage for rows_written/cold start, I'd rather see it integrated into the existing test infrastructure (see existing perf tests). Let's close this for now. In the future, we may want to expand the existing perf tests to also cover the CF adapter, which would be a better fit. Thanks for the work you did on it!

PR #1113, opened 2026-03-21
chore(nix): follow effect-utils nixpkgs authority

schickling-assistant:
## Summary
- make root `nixpkgs` follow `effect-utils/nixpkgs` in `devenv.yaml`
- refresh lock state to effect-utils strict-follow PR #428
- keep the downstream effect-utils build graph aligned across lock domains

## Rationale
Prepared pnpm trees coming from `effect-utils` are content-addressed against the upstream build graph. This repo should inherit that authority instead of overriding `effect-utils` to an ambient root pin.

## Validation
- `git diff --check`
- `devenv update effect-utils --override-input effect-utils git+https://github.com/overengineeringstudio/effect-utils?ref=schickling/2026-03-21-effect-utils-upstream-nixpkgs-authority`

Acting on behalf of the user.


schickling-assistant:
Depends on overengineeringstudio/effect-utils#428.

This branch intentionally points `effect-utils` at the upstream branch using the canonical `github:` input form so the downstream locks stay on the GitHub flake path instead of the raw git path that was breaking Devenv CI. After #428 lands, retarget the input back to the normal upstream ref before merging here.

Acting on behalf of the user.

PR #1125, opened 2026-03-24
fix(devenv): use effect-utils pinned pnpm 11

schickling-assistant:
## Summary

- Switches from `pkgs.pnpm` (nixpkgs, currently pnpm 10.x) to `effectUtils.lib.mkPnpm` which provides a pinned pnpm 11 build
- This ensures the livestore repo uses the same pnpm 11 version pinned in effect-utils, which is required for GVS-in-CI support

## Dependencies

- Depends on effect-utils PR https://github.com/overengineeringstudio/effect-utils/pull/451


schickling-assistant:
> Created on behalf of @schickling by Claude Code

Issue #1124, opened 2026-03-24
Improve error handling in `ClientSessionSyncProcessor`


IGassmann:
## Summary

`ClientSessionSyncProcessor` has three error handling gaps relative to our [error handling guidelines](../blob/dev/Users/igassmann/Downloads/effect-error-handling.md). Addressing these will improve observability, runtime correctness, and alignment with Effect conventions.

### 1. Background push fiber can die silently (highest priority)

**File:** `packages/@livestore/common/src/sync/ClientSessionSyncProcessor.ts` (lines 217–225)

The background push loop only recovers `RejectedPushError`. Any other failure kills the fiber via `tapCauseLogPretty` without triggering session shutdown or surfacing a processor-level error. This leaves the processor in a half-alive state: it can still pull events from the leader but can never push again.

This is the biggest gap — an operational failure that is neither modeled as a typed error nor escalated through a consistent fail-fast path.

**Expected:** Non-`RejectedPushError` failures in the push loop should either shut down the session (matching the pull loop's behavior) or be wrapped in a dedicated processor-level error and routed through the same shutdown path.

PR #1147, opened 2026-03-31
refactor: improve worker communication type safety by surfacing real error types

IGassmann:
## Problem

Every worker handler error — whether a SQLite failure, schema decode issue, rejected push, or unexpected crash — was collapsed into the same `UnknownError` wrapper before crossing the wire. This made it impossible to tell from the client side what actually went wrong. Additionally, the `runInWorker` / `runInWorkerStream` helpers relied on conditional return types (`TReq extends Schema.WithResult<infer A, ...> ? ... : never`) that required `as any` casts to satisfy the compiler, undermining the type safety they were meant to provide.

## Solution

Three-part approach, built up incrementally:

1. **Simplify generic signatures** — Replace the conditional-return-type pattern with a straightforward intersection constraint (`req: WorkerSchema.X & Schema.WithResult<A, I, E, EI, R>`), eliminating all `as any` / `as unknown as` casts in `runInWorker`, `runInWorkerStream`, `forwardRequest`, and `forwardRequestStream`.

2. **Separate transport errors from application errors** — Add an `isWorkerTransportError` type guard that identifies `WorkerError` and `ParseError` as infrastructure failures. Use `Effect.catchIf` / `Stream.refineOrDie` at the call site to convert these to defects instead of wrapping everything in `UnknownError`. These transport errors were already unrecoverable before — they were just being disguised as `UnknownError`. Now they surface as defects with their original identity preserved, which is more honest and easier to diagnose.

3. **Declare honest failure types** — Change worker schema requests from `failure: UnknownError` to `failure: Schema.Never` (or the real error, e.g. `RejectedPushError` for push). Remove `UnknownError.mapToUnknownError` from handler implementations. Update `ClientSessionLeaderThreadProxy` and `LeaderSyncProcessor` interfaces to reflect the narrower error channels.

Also replaces `timeoutDie`/`timeoutDieMsg` with `timeoutOrDie`/`timeoutOrDieMessage` that use `Effect.timeoutFailCause` to raise `TimeoutException` as a defect, keeping the error channel clean.

## Validation

`dt ts:check` and `dt lint:full` should pass. The changes are type-level and error-handling refactors with no behavioral changes to the happy path.

## Related issues

- https://github.com/livestorejs/livestore/issues/1124


packages/@livestore/adapter-expo/src/index.ts
```diff
@@ -309,8 +309,7 @@ const makeLeaderThread = ({
               .push(
                 batch.map((item) => new LiveStoreEvent.Client.EncodedWithMeta(item)),
                 { waitForProcessing: true },
-              )
-              .pipe(Effect.provide(layer), Effect.scoped),
+              ),
```
IGassmann, line 312: `Effect.provide(layer)` merges the layer's error type into the effect's error channel. Since `makeLeaderThreadLayer` returns `Layer.Layer<LeaderThreadCtx, UnknownError, ...>`, providing it adds `UnknownError` to the resulting effect. Previously the interface allowed `Effect.Effect<void, UnknownError | RejectedPushError>` so this was fine. Now that the interface narrowed to `Effect.Effect<void, RejectedPushError>`, keeping `Effect.provide(layer)` would produce `Effect.Effect<void, RejectedPushError | UnknownError>` which no longer satisfies the type. The provide was always a no-op at runtime (`syncProcessor.push` has `R = never` — it only uses closed-over values), so the removal is correct.

schickling (approved):

Issue #1111, opened 2026-03-20
Make retry policies of sync processors configurable


IGassmann:
## Problem

The retry policies used across sync processors are mostly hardcoded, making it difficult for consumers to tune retry behavior for their specific use cases.

### Current state

| Component | Scope | Strategy | Configurable |
|-----------|-------|----------|:---:|
| `LeaderSyncProcessor` (pull) | Offline reconnect | Infinite retry (default schedule) | No |
| `LeaderSyncProcessor` (push) | Push events | Exponential 1s → 30s cap, indefinite | No |
| `sync-cf` WS RPC | Socket open | Fixed 1s interval | No |
| `webmesh` WS edge | Socket open | Exponential 100ms ∪ fixed 5s | No |
| `webmesh` proxy channel | Send | Exponential 10ms base | No |
| `webmesh` direct channel | Handshake | 2 retries, 10ms timeout | No |
| Adapters (CF, Expo, Node, Web) | DB/OPFS reset | `exponentialBackoff10Sec` | No |
| `sync-s2` | Pull & push | 2 retries, 100ms spaced | **Yes** |
| `utils` RPC client | Socket transport | Configurable via options | **Yes** |

Only the S2 sync provider and the low-level RPC client layer expose retry configuration. The core `LeaderSyncProcessor` — which orchestrates all pull/push retry logic — has no configuration surface at all.

### Key hardcoded values

- **`LeaderSyncProcessor` push:** `Schedule.exponential(Duration.seconds(1))` capped at 30s — good default but not tunable.
- **`sync-cf` WS:** `retryTransientErrors: Schedule.fixed(1000)` — 1s fixed interval with no backoff or cap.
- **`webmesh` WS edge:** union of `Schedule.exponential(100)` and `Schedule.spaced(5000)` — complex strategy with no escape hatch.
- **Adapter reset ops:** all share `exponentialBackoff10Sec` (10ms base, factor 4, then 1s spaced, max 10s total).

## Proposal

Expose retry schedule configuration for at least the `LeaderSyncProcessor` pull and push paths, following the pattern already established by `sync-s2`:

```ts
// Strawman API — details TBD
{
  retry?: {
    pull?: Schedule.Schedule<...>
    push?: Schedule.Schedule<...>
  }
}
```

### Considerations

- The `LeaderSyncProcessor` is the highest-leverage target since it wraps all sync backend interactions.
[...]

PR #1149, opened 2026-03-31
feat(sync): add connectionStatus to NetworkStatus for reconnection UI

bohdanbirdie:
## Summary

- Adds `connectionStatus: 'connected' | 'disconnected' | 'reconnecting'` to `NetworkStatus` alongside existing `isConnected` boolean
- Non-breaking for consumers — `isConnected` stays as-is, `connectionStatus` is an additional field

## Problem

Previously, `store.networkStatus` only exposed `isConnected: boolean`. When a WebSocket connection dropped and the client was actively retrying with exponential backoff, the status showed `false` — identical to a fully disconnected state. Apps had no way to distinguish "offline and retrying" from "offline and idle", making it impossible to show meaningful reconnection indicators in the UI (e.g., "Reconnecting..." vs "Disconnected").

Linear example:
CleanShot 2026-04-01 at 14 31 13@2x

## How it works

- `RpcClient` sets `'reconnecting'` on transient socket errors (`Open`/`OpenTimeout`) when retry is configured, `'disconnected'` otherwise, `'connected'` on successful pong
- `connectionStatus` is the single source of truth — `isConnected` is derived from it in the aggregation layer to keep both fields in sync
- Backends without reconnection logic (HTTP, DO, Electric, S2) mirror `isConnected` — always `'connected'` or `'disconnected'`

## Known limitations

- No integration test for the `'reconnecting'` state through `RpcClient` — would require a mock socket layer. Unit test covers the state propagation through `makeNetworkStatusSubscribable`.

## Ideas for follow-up

- React `useNetworkStatus` hook (similar to existing `useSyncStatus`)
- Usage examples in example projects (e.g., showing a reconnection banner in todomvc)

Ref: #1111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -297,6 +297,8 @@ Key improvements include streaming pull operations (faster initial sync), a two-
 - Resilience: Improve sync provider robustness and align test helpers for CI and local development (#682, #646).
 - Header forwarding: Added `forwardHeaders` option to `makeDurableObject()` for cookie-based authentication. Headers are stored in WebSocket attachments to survive hibernation and accessible via `context.headers` in `onPush`/`onPull` callbacks (#929).
 - **Backend reset detection:** LiveStore now detects when a sync backend has been reset and handles it based on the `onBackendIdMismatch` option in `SyncOptions`. Default behaviour (`'reset'`) clears local storage and shuts down so the app can restart with fresh data. Alternative modes include `'shutdown'` (shut down without clearing) and `'ignore'` (continue with stale data). See [Backend Reset Detection docs](https://livestore.dev/building-with-livestore/syncing#backend-reset-detection) (#980).
+- **WebSocket retry backoff:** Replaced fixed 1-second retry with exponential backoff (1s to 30s cap, jittered) to reduce unnecessary requests during extended outages (#1111).
+- **Connection status:** `store.networkStatus` now includes a `connectionStatus` field (`'connected'`, `'disconnected'`, or `'reconnecting'`) so apps can distinguish active reconnection attempts from a fully disconnected state (#1111).
 
 ##### S2 sync backend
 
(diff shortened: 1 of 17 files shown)


IGassmann:
Thanks for the PR!

I'm trying to understand the practical value of the `reconnecting` state as a distinct status. With the current implementation, the WS backend retries forever. So once a connection drops, the client enters `reconnecting` and stays there indefinitely until it either succeeds or the tab closes — it never transitions to `disconnected`. While the non-WS backends never produce `reconnecting` at all.

So in practice, `reconnecting` seems equivalent to `isConnected: false` on the WS transport — every disconnection goes straight to `reconnecting` and stays there.

Interestingly, Linear itself doesn't use a tri-state model for this. If you inspect their `<AppStatusBadge>` component you can see that it contains 8 possible core states:

| State | Shown as | Meaning |
|---|---|---|
| `hidden` | *(nothing)* | Everything is fine |
| `bootstrapping` | "Initializing" | App is first loading |
| `syncing` | "Syncing N" | Pushing local changes to backend |
| `userOffline` | "Offline" | Your device has no internet |
| `deemedOffline` | "Offline" | You have internet, but Linear's backend is unreachable |
| `downtimeOffline` | "Offline" | Linear is down for maintenance |
| `downtimeReadonly` | "Read-only" | Maintenance mode, changes won't save |
| `saving` | "Caching" | Persisting data to local storage |


The displayed "Reconnecting..." state is a temporary local UI state that is shown only in `userOffline` state[^1] and  when the user has clicked on the badge. That badge click triggers `refreshOnlineStatus()` (internally driven by `navigator.onLine`) to re-check the connection. The state resets after 2 seconds. The distinction is purely a UI/timing concern, not a connection-layer status.

[^1]: The other offline-like states (`deemedOffline`, `downtimeOffline`) do not trigger this message because they are server-side problems that cannot be fixed by user interaction.

What's the concrete use case/end-user problem you're solving for where `connected` / `disconnected` isn't sufficient? Is there a scenario in your app where the UI would behave differently for `reconnecting` vs `disconnected`?

bohdanbirdie:
You might be right here
The change I made in https://github.com/livestorejs/livestore/pull/1144 kinda solved the problem I had. My local patch make the reconnect eventually stop, but backoff with forever retry is a fine default, so there is no need in third option now. PR lost it's value, closing it down
Thanks for the review and feedback!

PR #1161, opened 2026-04-07
fix: use `deepEqual` for event args comparison to ignore key order

### LINKED ISSUE #1160: [Bug]: Event equality fails when `args` key order is different
### Bug Description

This bit me this weekend. I was using psql db with `jsonb` column to store event `args`. I had syncing bug causing duplicate events. It turned out out equality check was not passing due to key order being different. Fix was to use `json` column which preserves order in the provided json.

https://github.com/livestorejs/livestore/blob/c74689bd6fc65e32e6391fb5d762ab4169cfd2de/packages/%40livestore/common/src/schema/LiveStoreEvent/client.ts#L188-L194

### Minimal Reproduction

N/A

### LiveStore Version

c74689bd6fc65e

@livestore/livestore

### Environment

```markdown
N/A
```

### Expected Behavior

Key order should not matter

### Actual Behavior

Events with args `{"b": 2, "a": 1}` and `{"a": 1, "b": 2}` are considered different events

### Error Logs/Screenshots

```shell

```

### Additional Context

_No response_

### Checklist

IGassmann:
## Problem

`isEqualEncoded` uses `JSON.stringify` to compare event args, which is key-order dependent. When a backend (e.g. PostgreSQL with `jsonb` columns) reorders object keys, semantically identical events like `{"a": 1, "b": 2}` and `{"b": 2, "a": 1}` are treated as different, causing duplicate events during sync.

## Solution

Replace `JSON.stringify(a.args) === JSON.stringify(b.args)` with `deepEqual(a.args, b.args)` from `@livestore/utils`, which performs recursive structural comparison regardless of key order. This utility is already used elsewhere in the codebase (reactive queries, React/Solid hooks).

The existing `TODO use schema equality here` comment is preserved for future consideration.

## Validation

Added 9 unit tests for `isEqualEncoded` covering:
- Identical args
- Different key order in flat args (was failing before fix)
- Different key order in nested args (was failing before fix)
- Different values (correctly not equal)
- Different keys (correctly not equal)
- Different event names (correctly not equal)
- Null args
- Array args
- Empty object args

```
 ✓ |@livestore/common| src/schema/LiveStoreEvent/client.test.ts (9 tests) 3ms
   Tests  9 passed (9)
```

## Related issues

- Closes #1160

### DIFF
--- a/packages/@livestore/common/src/schema/LiveStoreEvent/client.ts
+++ b/packages/@livestore/common/src/schema/LiveStoreEvent/client.ts
@@ -1,4 +1,4 @@
-import { memoizeByRef } from '@livestore/utils'
+import { deepEqual, memoizeByRef } from '@livestore/utils'
 import { Option, Schema } from '@livestore/utils/effect'
 
 import type { EventDef } from '../EventDef/mod.ts'
@@ -194,7 +194,7 @@ export const isEqualEncoded = (a: Encoded, b: Encoded) =>
   a.name === b.name &&
   a.clientId === b.clientId &&
   a.sessionId === b.sessionId &&
-  JSON.stringify(a.args) === JSON.stringify(b.args) // TODO use schema equality here
+  deepEqual(a.args, b.args) // TODO use schema equality here
 
 /**
  * Creates an Effect Schema union for all event types in a schema (with composite sequence numbers).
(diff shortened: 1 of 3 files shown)


schickling (approved):

packages/@livestore/common/src/schema/LiveStoreEvent/client.ts
```diff
@@ -194,7 +194,7 @@ export const isEqualEncoded = (a: Encoded, b: Encoded) =>
   a.name === b.name &&
   a.clientId === b.clientId &&
   a.sessionId === b.sessionId &&
-  JSON.stringify(a.args) === JSON.stringify(b.args) // TODO use schema equality here
+  deepEqual(a.args, b.args) // TODO use schema equality here
```
IGassmann, line 197: Benchmarked `deepEqual` (fast-deep-equal) vs `JSON.stringify` comparison across 5 scenarios (small args, different key order, early mismatch, nested args, 20-key flat args). Both land around 26–28M ops/sec — `deepEqual` is 2–5% faster in 4 of 5 cases, `JSON.stringify` is ~6% faster for nested objects. Effectively identical performance, so no concern here.

PR #1163, opened 2026-04-07
fix(rpc): patch @effect/rpc to use msgpackr/index-no-eval for CF Workers

bohdanbirdie:
## Problem

Durable Object-to-Durable Object sync via `@effect/rpc` silently fails when an RPC message contains 4+ objects with the same key structure (e.g., an array of `{_tag, requestId, exit}` objects) on deployed Cloudflare Workers.

### How it works

`@effect/rpc` uses `msgpackr` for its `RpcSerialization.msgPack` serializer. When packing an array of objects with identical keys, msgpackr encodes the first object as a **structure definition** and subsequent objects as **bare record IDs** referencing that definition.

On the decode side (`unpackMultiple`), a shared `readObject` function tracks how many times each structure is read. After the count exceeds 2, msgpackr tries to compile a fast reader via `new Function()`:

```
Object 1 → structure definition → createStructureReader() → readObject.count = 1   ✅
Object 2 → bare record ID 0x40 → readObject() → count = 2                         ✅
Object 3 → bare record ID 0x40 → readObject() → count = 3 → EXCEEDS threshold →
    new Function('r', 'return function(){return {_tag:r(),requestId:r(),...}}')
    → BLOCKED by CF Workers CSP                                                    ❌
```

The error is silently swallowed by `@effect/rpc`'s catch-all in the decode path (`catch { return [] }`), causing the server to return empty responses.

### Prerequisites

- **`allow_eval_during_startup`** compat flag (default since `compatibility_date >= 2025-06-01`). This allows msgpackr's startup probe `try { new Function('') }` to succeed, keeping JIT enabled. Without this flag, the probe fails at startup and JIT is globally disabled.
- **4+ same-structure objects in a single `pack()` call.** Within one `pack()`, msgpackr uses structure records — the first object defines the structure, subsequent objects use bare record IDs. The `readObject.count` accumulates within a single `unpackMultiple` call.

### Reproduction

**Must deploy to actual CF Workers** — miniflare does not enforce the `new Function()` restriction locally.

## Fix

Patch `@effect/rpc` to import from `msgpackr/index-no-eval` instead of `msgpackr`. This is msgpackr's official entry point for CSP-restricted environments — it strips all `new Function()` calls at build time, falling back to the interpreted decode path. The record/structure optimization is preserved (payloads stay compact), only the JIT compilation step is removed.

The patch is minimal — a single-line import change in each of the ESM and CJS builds.

Upstream PR: https://github.com/Effect-TS/effect/pull/6161 — once landed, this patch can be removed.

### DIFF
--- a/patches/@effect__rpc@0.75.0.patch
+++ b/patches/@effect__rpc@0.75.0.patch
@@ -1,0 +1,26 @@
+diff --git a/dist/cjs/RpcSerialization.js b/dist/cjs/RpcSerialization.js
+index 43c28ee14b8167c4ceb1a9ec59dfa7d4ad0d3b64..1d0f0c2218a46f3816c43d9cffadefb700629332 100644
+--- a/dist/cjs/RpcSerialization.js
++++ b/dist/cjs/RpcSerialization.js
+@@ -7,7 +7,7 @@ exports.ndjson = exports.ndJsonRpc = exports.msgPack = exports.layerNdjson = exp
+ var Context = _interopRequireWildcard(require("effect/Context"));
+ var Layer = _interopRequireWildcard(require("effect/Layer"));
+ var _Predicate = require("effect/Predicate");
+-var Msgpackr = _interopRequireWildcard(require("msgpackr"));
++var Msgpackr = _interopRequireWildcard(require("msgpackr/index-no-eval"));
+ function _interopRequireWildcard(e, t) { if ("function" == typeof WeakMap) var r = new WeakMap(), n = new WeakMap(); return (_interopRequireWildcard = function (e, t) { if (!t && e && e.__esModule) return e; var o, i, f = { __proto__: null, default: e }; if (null === e || "object" != typeof e && "function" != typeof e) return f; if (o = t ? n : r) { if (o.has(e)) return o.get(e); o.set(e, f); } for (const t in e) "default" !== t && {}.hasOwnProperty.call(e, t) && ((i = (o = Object.defineProperty) && Object.getOwnPropertyDescriptor(e, t)) && (i.get || i.set) ? o(f, t, i) : f[t] = e[t]); return f; })(e, t); }
+ /**
+  * @since 1.0.0
+diff --git a/dist/esm/RpcSerialization.js b/dist/esm/RpcSerialization.js
+index a28fe2fc7bced37c01657e4664c63b56fcac1c81..483e4c48db9e8212e21b0cc3af25c071818a72e5 100644
+--- a/dist/esm/RpcSerialization.js
++++ b/dist/esm/RpcSerialization.js
+@@ -4,7 +4,7 @@
+ import * as Context from "effect/Context";
+ import * as Layer from "effect/Layer";
+ import { hasProperty } from "effect/Predicate";
+-import * as Msgpackr from "msgpackr";
++import * as Msgpackr from "msgpackr/index-no-eval";
+ /**
+  * @since 1.0.0
+  * @category serialization
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -214,6 +214,9 @@ overrides:
 
 packageExtensionsChecksum: sha256-7YcoS1641lRAmSqRcs4IGptEje/ekJAQXXTonMs6mHQ=
 
+patchedDependencies:
+  '@effect/rpc@0.75.0': 509c7105da5cdad85d26339dc08329ab85898017d47be8f17b8243ae645e3c4d
+
 importers:
 
   .: {}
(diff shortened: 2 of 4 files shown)


bohdanbirdie:
**Upstream issue filed:** https://github.com/Effect-TS/effect/pull/6161

This PR proposes a fix in `@effect/rpc`'s `RpcSerialization.msgPack` to pass `{ useRecords: false }` to msgpackr's `Packr`/`Unpackr` constructors, preventing the JIT `new Function()` path that is blocked by CF Workers CSP.

If the upstream fix lands, the pnpm patch in this PR can be removed.

IGassmann (commented):
DELETED

IGassmann (changes requested):
### Patch approach: prefer `msgpackr/index-no-eval`

The current patch disables msgpackr's record/structure optimization entirely via `{ useRecords: false }` and adds an unexplained `int64AsType: 'number'` option. A cleaner approach is to use msgpackr's official `index-no-eval` entry point (available since [v1.8.0](https://github.com/kriszyp/msgpackr/issues/86)), which strips all `new Function()` code from the module at build time:

```diff
-import * as Msgpackr from "msgpackr";
+import * as Msgpackr from "msgpackr/index-no-eval";
```

Advantages:
- **Uses the officially supported solution** from the msgpackr maintainer for CSP-restricted environments
- **Doesn't disable the record/structure optimization** — payloads stay compact, the decode just falls back to the interpreted path instead of trying to JIT compile
- **Simpler patch** — only changes import paths, no constructor argument modifications
- **Removes the unexplained `int64AsType: 'number'`** addition which seems to be an unrelated behavior change.

I verified that the `index-no-eval` build:
- Exports the same `Packr`/`Unpackr` API (drop-in replacement)
- Contains **zero** `new Function()` calls (vs 2 in the regular build)
- Round-trips correctly, including cross-compat with the regular build's output
- Produces a CF Worker bundle with **zero** `new Function()` calls

### Scope

This PR bundles four independent changes. I'd suggest splitting:

1. **The `@effect/rpc` msgpackr patch** (the core fix) — this PR
2. **`packages/@livestore/common-cf/src/do-rpc/client.ts` `flat(1)` change** — separate PR. This fixes a different bug (CF merging multiple `enqueue()` calls into a single `read()`) that is independent of the msgpackr JIT issue.

...and discarding:

3. **`examples/cf-ai-research-agent/`** — we want to avoid unnecessarily adding more example apps to maintain.
4. **`pnpm-workspace.yaml`** — includes unrelated edits (alphabetizing `overrides`, adding new `allowBuilds`, reformatting `supportedArchitectures`).

IGassmann (approved):

acusti:
this caught my attention when i saw it involves the sync DOs and `msgpackr`. i ran into overlapping issues when i switched my entire data model over to LiveStore and i have a local patch that replaces `RpcSerialization.msgPack.unsafeMake()` with a custom `Packr`/`Unpackr` setup configured with `useRecords: false`, which avoids the record/JIT path that breaks under Cloudflare Workers CSP. seems like that part has been superseded by this PR, which is great.

that said, my patch also includes a separate streaming fix in `common-cf` that buffers partial MessagePack data across `ReadableStream` reads, uses `unpackMultiple()` to decode incrementally, and surfaces a truncated-stream error if bytes are still pending at EOF. should i put together a minimal PR just for that remaining piece?

bohdanbirdie:
Update: I think that current PR only affected the livestore monorepo but not the consumers of livestore
I still had to add a patch to Effect in my project, it's an oversight from my side but not yet sure what is a right way to fix this

Best scenario is if the PR in effect will be merged

acusti:
@bohdanbirdie i put together a branch with an alternative approach to the `RpcSerialization.msgPack.unsafeMake()` usage in `@livestore/common-cf`. i’m going to open a PR from it to show an approach that avoids the eval-related workerd CSP incompatibility and includes the stream-boundary fix i mentioned above.

acusti:
@bohdanbirdie see #1170

acusti:
@IGassmann in your review, you suggest using `import * as Msgpackr from "msgpackr/index-no-eval";`, so i did that in my first version of #1170, but that leads to this type error:

```
packages/@livestore/common-cf/src/do-rpc/msgpack.ts(1,32): error TS7016: Could not find a declaration file for module 'msgpackr/index-no-eval'. '/home/runner/work/livestore/livestore/.pnpm-home/store/v11/links/@/msgpackr/1.11.9/346944ca1495d7bbbb1afeffa74819c8661437de196c9945d5867e4909357a6b/node_modules/msgpackr/dist/index-no-eval.cjs' implicitly has an 'any' type.
Try `npm i --save-dev @types/msgpackr` if it exists or add a new declaration (.d.ts) file containing `declare module 'msgpackr/index-no-eval';`
```

i can add a `.d.ts` file for `msgpackr/index-no-eval` (`msgpackr` ships with types and i don’t see any `@types/msgpackr` package so i’m pretty sure that’s not the right solution) if that seems best to you. though the patch i have been using imports from `msgpackr` with the `{ mapsAsObjects: true, useRecords: false }` options and fixes the issue for me in my sync DO instances in prod.

bohdanbirdie:
@acusti I reviewed this option too, but decided that extracting code from effect and owning it in livestore is an extra liability 
It's best if the thing is fixed in effect and perhaps with local patch
But let's see if Igor agrees

IGassmann:
@bohdanbirdie @acusti, `msgpackr` has implemented a [fix](https://github.com/kriszyp/msgpackr/commit/c6e831d5c4af4b05824a72d351ef4de34d51b43c). Whenever, it's shipped we should be able to include it in LiveStore by using a pnpm override.

IGassmann:
The upstream fix has landed — msgpackr 1.11.10 wraps the JIT `new Function()` in a try/catch, falling back to the interpreted path. I've opened https://github.com/Effect-TS/effect/pull/6191 to bump `@effect/rpc`'s minimum `msgpackr` version to `^1.11.10`. Once that's released, this patch can be dropped.

IGassmann:
Update: the monorepo-only patch has been reverted in #1178 since it didn't propagate to downstream consumers.

Tracking issue: #1179 — includes a simpler workaround for downstream users using a pnpm [`overrides`](https://pnpm.io/settings#overrides):

```json
{
  "pnpm": {
    "overrides": {
      "msgpackr": ">=1.11.10"
    }
  }
}
```

This forces pnpm to resolve `msgpackr` to a version that includes the CF Workers fix ([kriszyp/msgpackr@c6e831d](https://github.com/kriszyp/msgpackr/commit/c6e831d5c4af4b05824a72d351ef4de34d51b43c)), without needing a patch file.

The permanent fix is blocked on Effect-TS/effect#6191 landing and `@effect/rpc` publishing a new version.

PR #1170, opened 2026-04-14
[codex] fix common-cf DO RPC msgpack handling on workers

acusti:
## Summary

- switch `@livestore/common-cf` DO RPC off `RpcSerialization.msgPack.unsafeMake()` and onto a direct `msgpackr` parser configuration that avoids the eval-dependent path in workerd
- buffer partial MessagePack payloads across `ReadableStream` reads and decode concatenated frames with `unpackMultiple()`
- add focused regression coverage for split and merged DO RPC stream frames

## Why

After [#1163](https://github.com/livestorejs/livestore/pull/1163), the monorepo has a workspace patch for `@effect/rpc`, but published consumers can still hit the same workerd CSP issue because `@livestore/common-cf` was still using `RpcSerialization.msgPack.unsafeMake()` directly.

This change makes the consumer-facing DO RPC path use a workers-safe parser directly in `common-cf` and fixes the related stream framing issue where a single MessagePack payload can be split across multiple `ReadableStream` reads.

## Root Cause

- Cloudflare workerd blocks the eval-based fast path used by the default msgpack setup
- the DO RPC streaming client assumed each `ReadableStream` read contained a complete MessagePack payload, which is not guaranteed when frames are split or concatenated by the transport

## Validation

- `pnpm --dir packages/@livestore/common-cf exec vitest run`

### DIFF
--- a/genie/external.ts
+++ b/genie/external.ts
@@ -99,6 +99,7 @@ export const livestoreOnlyCatalog = {
   'qrcode-generator': '1.4.4',
   '@iarna/toml': '3.0.0',
   '@graphql-typed-document-node/core': '3.2.0',
+  msgpackr: '1.11.9',
   'astro-expressive-code': '0.41.5',
   'expressive-code': '0.41.5',
   'expressive-code-twoslash': '0.5.3',
--- a/packages/@livestore/common-cf/package.json
+++ b/packages/@livestore/common-cf/package.json
@@ -30,7 +30,8 @@
   },
   "dependencies": {
     "@cloudflare/workers-types": "4.20251118.0",
-    "@livestore/utils": "workspace:^"
+    "@livestore/utils": "workspace:^",
+    "msgpackr": "1.11.9"
   },
   "devDependencies": {
     "@livestore/utils-dev": "workspace:^",
--- a/packages/@livestore/common-cf/package.json.genie.ts
+++ b/packages/@livestore/common-cf/package.json.genie.ts
@@ -12,7 +12,7 @@ const runtimeDeps = catalog.compose({
   workspace: workspaceMember('packages/@livestore/common-cf'),
   dependencies: {
     workspace: [utilsPkg],
-    external: catalog.pick('@cloudflare/workers-types'),
+    external: catalog.pick('@cloudflare/workers-types', 'msgpackr'),
   },
   devDependencies: {
     workspace: [utilsDevPkg],
--- a/packages/@livestore/common-cf/src/do-rpc/client.ts
+++ b/packages/@livestore/common-cf/src/do-rpc/client.ts
@@ -1,15 +1,7 @@
-import {
-  Effect,
-  Fiber,
-  FiberMap,
-  Layer,
-  RpcClient,
-  type RpcMessage,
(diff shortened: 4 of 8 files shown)


acusti:
i got all of the type errors and lint issues resolved. i should’ve kept the PR as a draft, but regardless, it’s ready for review now.

IGassmann (changes requested):
Thanks for the fix and especially the regression tests for stream framing.

### Split into two PRs

This PR bundles two independent fixes:

1. **Eval fix** — replacing `RpcSerialization.msgPack.unsafeMake()` with a direct `msgpackr` parser configured with `{ useRecords: false }`
2. **Stream framing fix** — buffering partial MessagePack payloads across `ReadableStream` reads with `unpackMultiple()`

These address different bugs with different root causes and carry different validation requirements and risk profiles — the eval fix is a straightforward config swap while the stream framing fix involves buffering state across reads and error handling at EOF. Separate PRs make review, bisection, and independent revert easier.

### Eval fix may not need a custom parser

Regarding the eval fix specifically: `msgpackr` has implemented an [upstream fix](https://github.com/kriszyp/msgpackr/commit/c6e831d5c4af4b05824a72d351ef4de34d51b43c) (https://github.com/livestorejs/livestore/pull/1163#issuecomment-4250620456). Once that's shipped, we should be able to resolve this with a pnpm override rather than adding `msgpackr` as a direct dependency and maintaining a custom parser wrapper. That may make the eval fix PR unnecessary — worth waiting to see how quickly the upstream release lands before adding the extra dependency and abstraction.

The stream framing fix stands on its own regardless, so that one can move forward independently.

acusti:
@IGassmann thanks for the review, and it’s great to see the upstream fix in `msgpackr`! wrt to the PR representing two independent fixes, the implementation of the fixes wound up being dependent on each other. i was originally gonna make two separate commits, but the one didn’t make that much since without the other. regardless, i will try to see if there’s a simpler version i can come up with based on only needing to address the stream framing fix.

IGassmann:
Update: we've opened a tracking issue at #1179 with a simpler workaround for downstream users — a pnpm [`overrides`](https://pnpm.io/settings#overrides) to force `msgpackr >=1.11.10`:

```json
{
  "pnpm": {
    "overrides": {
      "msgpackr": ">=1.11.10"
    }
  }
}
```

This avoids needing a custom msgpackr setup or a patch file. The permanent fix is blocked on Effect-TS/effect#6191.

IGassmann:
Closing this — the eval fix is now fully resolved upstream. `msgpackr` 1.11.10 wraps the JIT `new Function()` call in a try/catch, and Effect has picked it up in `@effect/rpc@0.75.1` / `@effect/platform@0.96.1` (Effect-TS/effect#6191). Once we [update our Effect dependencies](https://github.com/livestorejs/livestore/issues/1179#issuecomment-4302507882), `RpcSerialization.msgPack.unsafeMake()` will work on Cloudflare Workers without a custom parser.

@acusti if the stream framing fix (buffering partial MessagePack payloads across `ReadableStream` reads) is still relevant, feel free to open a separate PR for that.

PR #1238, opened 2026-05-13
refactor(sync): use `SubscriptionRef` for `syncStateRef`

IGassmann:
## Problem

`ClientSessionSyncProcessor` tracked the current sync state in a hand-rolled mutable `syncStateRef = { current: SyncState }` wrapper and notified observers via a separate `syncStateUpdateQueue: Queue<SyncState>`. Every write had to update both, and on the pull path the cached value was mutated *before* materialization completed, so a subscriber reading the ref could see a state the observability stream had not yet announced.

## Solution

Collapse both into a single `SubscriptionRef<SyncState>`.

- One source of truth instead of a ref/queue dual-write that has to be kept in sync by hand.
- Subscribers and cached reads cannot disagree mid-step: both the pull and push merges run inside `SubscriptionRef.updateEffect`, so the new state is only published after materialization completes (pull) and after the new events have been offered to the leader push queue (push).
- New subscribers receive the current value automatically via `SubscriptionRef.changes`, so `Store.syncStatusStream()` no longer needs to concatenate an initial value.

The internal interface field changes from `Subscribable.Subscribable<SyncState>` to `SubscriptionRef.SubscriptionRef<SyncState>`. The processor is reached through `StoreInternalsSymbol`, so this is not a public API change.

## Validation

- \`dt ts:check\` passes.
- Behavior is covered by the existing client-session sync integration tests.

## Related issues

- none

### DIFF
--- a/.changeset/refactor-sync-state-subscription-ref.md
+++ b/.changeset/refactor-sync-state-subscription-ref.md
@@ -1,0 +1,6 @@
+---
+"@livestore/common": patch
+"@livestore/livestore": patch
+---
+
+Internal: back `ClientSessionSyncProcessor` sync state with a `SubscriptionRef` (#1238).
--- a/packages/@livestore/common/src/sync/ClientSessionSyncProcessor.ts
+++ b/packages/@livestore/common/src/sync/ClientSessionSyncProcessor.ts
@@ -6,11 +6,11 @@ import {
   Exit,
   FiberHandle,
   Option,
-  Queue,
   Schema,
   type Scope,
   Stream,
   Subscribable,
+  SubscriptionRef,
 } from '@livestore/utils/effect'
 
 import type { ClientSession } from '../adapter-types.ts'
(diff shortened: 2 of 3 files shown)


IGassmann:
Closing this PR: the refactor is fundamentally incompatible with the synchronous `Store.commit` path.

## What CI surfaced

`build-and-deploy-examples-src` → `Test examples` fails three `cf-chat` Playwright specs because the app's WebServer crashes during `Store.commit`:

```
(FiberFailure) AsyncFiberException: Fiber #… cannot be resolved synchronously.
This is caused by using runSync on an effect that performs async work
    at client-session-sync-processor:push (.../ClientSessionSyncProcessor.ts:307:59)
    at LiveStore:commit (.../store.ts:841:14)
```

Reproduced locally:

```sh
cd examples/cf-chat
PLAYWRIGHT_BROWSERS_PATH=… CI=1 pnpm exec playwright test --grep "bot reacts to messages"
```

Same `AsyncFiberException` on first commit.

## Root cause

`Store.commit` runs its body via `Runtime.runSync` (`store.ts:859`). Inside, it `yield*`s `syncProcessor.push(...)`.

This PR routes every state transition through `SubscriptionRef` (`set` in the first commit, `updateEffect` in the second). In Effect, **every `SubscriptionRef` write — including `set`, `update`, and `updateEffect` — runs inside `semaphore.withPermits(1)`**. That semaphore introduces an async suspension regardless of contention, so the previously-synchronous `commit` Effect can no longer be resolved by `runSync`.

Pre-refactor `push` did `syncStateRef.current = newState` + `BucketQueue.offer` — synchronous in fiber terms, which is why `runSync` worked. The unit/integration suites passed in CI because they use `runFork`/`runPromise`; only the example apps exercise the sync `commit` path.

## Why this can't be patched in place

- Switching `updateEffect` back to `set` doesn't help — `set` also goes through the semaphore.
- Keeping the old mutable `syncStateRef` for the synchronous write and publishing to a `SubscriptionRef` for observers reintroduces the dual-write the PR was designed to eliminate, defeating the stated motivation.

PR #1264, opened 2026-06-01
Fix DO-RPC streaming truncation by returning the stream-runner promise from ReadableStream.start()

bohdanbirdie:
## Problem

When a Durable Object acts as a LiveStore client over the **DO-RPC transport**, multi-chunk streaming responses (e.g. a cold-boot catchup pull) are silently truncated to the first chunk. The client receives only the head of the payload, its sync head never advances, and synchronization stalls permanently.

The root cause is in `@livestore/common-cf`'s `do-rpc/server.ts`. `createStreamingResponse` runs the Effect stream runner **fire-and-forget** inside `ReadableStream.start()`:

```ts
// Run the stream processing
runStream.pipe(Effect.provide(layer), Effect.scoped, Effect.tapCauseLogPretty, Effect.runPromise)
```

The Web Streams contract lets `start()` return a promise, and the Cloudflare runtime relies on it to know when the stream is done producing chunks. Because the promise is dropped, CF can tear the stream down before the 2nd+ `controller.enqueue()` calls fire — dropping the tail of any multi-chunk payload.

## Solution

Return the promise from `start()` so the runtime keeps the stream open until every chunk is enqueued and `controller.close()` is called:

```ts
return runStream.pipe(Effect.provide(layer), Effect.scoped, Effect.tapCauseLogPretty, Effect.runPromise)
```

One-line behavioural fix plus an explanatory comment. No API change; preserves the existing `.flat(1)` merged-enqueue normalization from #1167.

This is the first of the two follow-ups @IGassmann invited when #1170 was closed and asked to be split into smaller PRs. It is the small, self-contained half (server framing). A separate follow-up will cover the client-side stream decoding.

## Validation

- Confirmed against the Web Streams `ReadableStream` underlying-source contract: an async `start()` returning a promise defers stream teardown until it settles.
- Behaviour matches the dist-level patch a downstream consumer (Cloudstash) has been running in production to unstick the stall.
- The existing DO-RPC streaming tests live in `packages/@livestore/common-cf/src/do-rpc/rpc.test.ts` (wrangler-based); CI exercises this path. _Note: the full wrangler suite was not run in my local sandbox (dev environment unavailable); the change is a single `return` and is covered by CI._

## Related issues

- #1170 — earlier combined attempt (closed); @IGassmann asked to split it and invited this follow-up.
- #1167 — `.flat(1)` merged-enqueue normalization (preserved here).
- #1163 — `msgpackr` → `msgpackr/index-no-eval` for CF Workers CSP (merged; unrelated to this framing fix but part of the same DO-RPC hardening effort).

### DIFF
--- a/.changeset/fix-do-rpc-server-stream-return.md
+++ b/.changeset/fix-do-rpc-server-stream-return.md
@@ -2,4 +2,4 @@
 "@livestore/common-cf": patch
 ---
 
-Fix Durable Object RPC streaming responses being truncated to the first chunk. `createStreamingResponse` now returns the stream-runner promise from `ReadableStream.start()`, so the Cloudflare runtime keeps the stream open until every chunk is enqueued and `close()` is called. Without this, the tail of any multi-chunk catchup payload could be dropped, permanently stalling a Durable-Object-as-LiveStore-client sync head (#1170).
+Fix Durable Object RPC streaming responses being truncated to the first chunk. `createStreamingResponse` now returns the stream-runner promise from `ReadableStream.start()`, so the Cloudflare runtime keeps the stream open until every chunk is enqueued and `close()` is called. Without this, the tail of any multi-chunk catchup payload could be dropped, permanently stalling a Durable-Object-as-LiveStore-client sync head (#1264, follow-up to #1170).
--- a/.changeset/fix-do-rpc-server-stream-return.md
+++ b/.changeset/fix-do-rpc-server-stream-return.md
@@ -1,0 +1,5 @@
+---
+"@livestore/common-cf": patch
+---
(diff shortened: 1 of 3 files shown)


bohdanbirdie:
It's unclear this is actually a real issue. I couldn't reproduce any effect from this `return` — including on a deployed Worker, where toggling it made no difference to a multi-chunk DO-RPC stream under natural draining, a mid-stream `ctx.abort()`, or a redeploy/runtime-roll. The cold-boot catchup stall this was meant to address turned out to be the client-side msgpack parser bug, fixed in #1266. Closing for now; can revisit if it resurfaces.

PR #1266, opened 2026-06-01
fix(common-cf): scope DO-RPC msgpack parser per request

bohdanbirdie:
## Problem

A Durable Object running as a LiveStore client over the **DO-RPC transport** can permanently stall its sync after a cold boot. It pulls the missing events as one **multi-chunk** catchup stream, the persisted sync head freezes below the eventlog head, and it never recovers — a destructive rebase can even shrink the local eventlog (data loss). Surfaced by a downstream app running a DO as a LiveStore client.

## Root cause

The DO-RPC client created **one** `RpcSerialization.msgPack` parser at connection scope and shared it across every `send`. That parser keeps a single stateful slot: when a buffer ends mid-frame it stashes the tail and prepends it on the next `decode()`.

Cloudflare DO RPC splits a streamed response into arbitrary ~4KB reads that don't align with msgpack frames, so a catchup frame straddles two reads. When a concurrent unary push-response decode runs on the shared parser **between** a straddling read and its continuation, it clears the stash — the straddling frame never reassembles, the rest of the catchup is dropped, and the head freezes.

It's a concurrency bug, not runtime-specific: a *single* stream self-heals (its reads carry the tail across uninterrupted), so isolated stream tests pass. It only trips when catchup-pull and push decodes interleave on the same parser — which reproduces in plain Node.

## Fix

Scope the parser per request/response, matching `RpcClient.makeProtocolHttp`. Each `send` now owns its parser, so the incomplete-frame buffer is never shared across concurrent calls. No new dependency, and the Effect serializer's incremental flush is preserved. Thanks @IGassmann for the suggestion.

## Tests

- **`stream-concurrency.test.ts`** (new, pure-Node, ~30 ms): drives the real `layerProtocolDurableObject` with a straddling catchup stream and one interleaved unary decode. Times out (frozen head) on the shared-parser code, passes with the per-send parser. Runs in CI via `mono test unit` — no workerd needed.
- `rpc.test.ts`: streaming/interruption coverage unchanged (still green).

## Related

- #1264 — the server-side half (returning the stream-runner promise). Independent of this fix.
- #1167 — `.flat(1)` merged-enqueue normalization (preserved here).


IGassmann (commented):
Why didn't you opt instead for scoping the existing Effect msgpack parser per DO RPC `send`, like `RpcClient.makeProtocolHttp` does, instead of adding raw `msgpackr` buffering here?

That would mean one `RpcSerialization.msgPack.unsafeMake()` per request/response, passed into `processReadableStream` for stream responses and used for the unary response decode. It seems like that would isolate the incomplete-frame buffer across concurrent DO RPC calls while preserving Effect’s current incremental flush behavior and avoiding a direct `msgpackr` dependency.

PR #1289, opened 2026-06-03
Detect npm latest vs release/version.json drift via daily health check

schickling-assistant:
## Problem

When a release PR merges but the npm publish job fails (token expired, OIDC misconfig, network flake), `release/version.json` on `main` advances past what npm `latest` actually serves. Today nothing surfaces that gap — the next release just rolls forward and downstream users keep installing the old version.

## Solution

Daily workflow at 09:00 Europe/Berlin (and `workflow_dispatch`) compares `npm view @livestore/livestore version` against the `.version` field in `release/version.json` on `main`.

Behaviour:

- pre-release file versions (e.g. `0.4.0-dev.26`) are skipped — `latest` only ever points at a stable release.
- inside a 48h grace window: `::notice::` with the elapsed delta.
- past 48h: opens (or daily-warms) a GitHub issue titled \`npm latest tag drifted from release/version.json for >48h\` with label \`bug\`.

\"First seen\" is derived from the GitHub Release tag's \`published_at\` when available, falling back to the commit time of \`release/version.json\` on \`main\`.

## Trade-offs / follow-ups

- The repo lacks the structured \`type:* area:* origin:*\` label taxonomy from CLAUDE.md, so the issue is filed with only the existing \`bug\` label. If the team later adds \`area:release\` / \`origin:janitor\`, swap them into \`ISSUE_LABELS\` in \`scripts/src/commands/health/npm-version-drift.ts\`.

## Validation

- \`bun scripts/src/commands/health/npm-version-drift.ts\` runs locally and currently emits the pre-release \`::notice::\` path (skipping the check) because \`release/version.json\` is a dev version.
- \`devenv tasks run lint:check\` passes.
- \`devenv tasks run genie:run --mode before --no-tui\` regenerates \`.github/workflows/health-npm-version-drift.yml\` cleanly.

## Files

- \`.github/workflows/health-npm-version-drift.yml.genie.ts\` — workflow source.
- \`scripts/src/commands/health/npm-version-drift.ts\` — the comparison + issue management.
- \`CHANGELOG.md\` — entry under \`### Internal Changes > Development Tooling\`.

### DIFF
--- a/.github/workflows/health-npm-version-drift.yml
+++ b/.github/workflows/health-npm-version-drift.yml
@@ -1,0 +1,31 @@
+# Generated file - DO NOT EDIT
+# Source: health-npm-version-drift.yml.genie.ts
+
+name: 'Health: npm version drift'
+
+on:
+  schedule:
+    - cron: 0 8 * * *
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: main
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+      - name: Compare npm latest vs release/version.json
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: bun scripts/src/commands/health/npm-version-drift.ts
--- a/.github/workflows/health-npm-version-drift.yml.genie.ts
+++ b/.github/workflows/health-npm-version-drift.yml.genie.ts
@@ -1,0 +1,45 @@
+import { defaultActionlintConfig, githubWorkflow } from '../../genie/repo.ts'
+
+/**
+ * Daily health check: compare npm `latest` for `@livestore/livestore` to the
+ * version recorded in `release/version.json` on `main`. Drift inside a 48h
+ * grace window is just a `::notice::`; sustained drift opens (or warms) a
+ * `type:bug area:release` issue.
+ *
+ * Intentionally minimal: no devenv / nix setup. Bun is already on the
+ * GitHub-hosted runner image and the script only needs `npm view`, `gh`, and
+ * `git log`.
+ */
+export default githubWorkflow({
+  name: 'Health: npm version drift',
+  actionlint: defaultActionlintConfig,
+
+  on: {
+    schedule: [{ cron: '0 8 * * *' }],
+    workflow_dispatch: {},
+  },
+
+  // Issue management requires write; everything else is read-only.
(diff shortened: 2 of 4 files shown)


schickling-assistant:
Closing in favour of #1504, which addresses the same failure mode at release time rather than from the outside a day later.

**Why supersede rather than land this**

The gap is real — a publish can succeed while `latest` keeps resolving to the previous release. But `release.ts` already verified every package post-publish (`npm view <pkg>@<version> version`, with retry); it just never checked the *dist-tag*. Adding that assertion to the existing loop covers all packages, in the same run, as a hard release failure, rather than one package up to 24h + 48h later as an issue.

**Two problems found while reviewing this branch**

1. **The check was fail-open — it could never fire.** `firstSeen` falls back to `versionFileLastCommitAt()`, which runs `git log -1 --format=%cI -- release/version.json`. The workflow checks out with no `fetch-depth`, i.e. a depth-1 shallow clone, where the path-limited log collapses to the tip commit:

   ```
   full clone,    real last change to release/version.json : 2026-06-02T08:08:44Z
   depth-1 clone, same command                             : 2026-07-15T09:35:03Z  (= tip)
   ```

   That timestamp is always recent on an active repo, so `hoursElapsed` is always below the 48h grace and the check only ever emits a `::notice::`. The fallback is reached whenever no GitHub Release exists for the file version — which is exactly the "publish job failed" case this was written for.

   The underlying design issue is `?? new Date()`: "can't determine when drift started → assume now → therefore fine". A monitor's unknown-timestamp path needs to fail closed.

2. **Single-package coverage.** Only `@livestore/livestore` was checked, so a partial publish — some packages tagged, some not — stayed invisible. That is both likelier and more damaging than a wholesale miss.

Also worth noting for anyone revisiting: `isPrerelease = (v) => v.includes('-')` is hand-rolled; the repo already depends on `semver` and `release.ts` parses prereleases properly.

Nothing here was wasted — this PR is what surfaced the dist-tag gap in the first place. #1504 carries it forward.


<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_name` | cl1-lagoon |
| `agent_session_id` | 4f0de924-6d8e-4779-9a7d-c30c8fa3d2a4 |
| `agent_tool` | Claude Code |
| `agent_tool_version` | 2.1.220 |
| `agent_runtime` | Claude Code 2.1.220 |
| `agent_model` | claude-opus-5 |
| `runtime_profile` | /nix/store/57hz15fbcg9c3hasm0ny3axd5pg33rh3-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/d9hvhhg1k9n3j20dgcch7sz3sg98l377-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | schickling |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@b785b9f |
</details>

PR #1292, opened 2026-06-03
docs(rfcs): RFC 0003 — Remove Client Document API

IGassmann:
Adds [RFC 0003 — Remove Client Document API](https://github.com/livestorejs/livestore/blob/igor/rfc-remove-client-docs/contributor-docs/rfcs/0003-remove-client-document-api.md)

### DIFF
--- a/contributor-docs/rfcs/0003-remove-client-document-api.md
+++ b/contributor-docs/rfcs/0003-remove-client-document-api.md
@@ -2,9 +2,7 @@
 
 ## Context
 
-LiveStore applications define state as SQLite tables and derive that state by committing events that are processed by materializers.
-
-For durable application data, users usually define:
+LiveStore applications define state as SQLite tables and derive that state by committing events that are processed by materializers:
 
 ```ts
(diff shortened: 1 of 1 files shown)


raymonddaikon:
I would lean towards path 1 or 3. I don't think client state events should bloat LiveStore. However it may be useful to persist a single LWW snapshot of certain client state to LiveStore; I'm thinking things like user settings. This way we can reuse the same sync infra without adding the overhead of events from high churn state where we don't care about the history.

IGassmann:
@raymonddaikon, thanks for jumping in!

Just to clarify: Path 3 as written and the current Client Document API are also event-backed, so for the same number of updates they produce the same number of client-only events as Path 2. The difference is semantic granularity, not event count.

Is your concern about bloating, event log growth, or did you mean something else?

raymonddaikon:
Definitely about event log growth, for user scoped data we don't necessarily need an event log, we can just force server authoritative state like a traditional db.

Issue #1217, opened 2026-05-05
Sync hash mismatch when event args use Schema.UndefinedOr or loose Schema.optional


IGassmann:
## Summary

Locally-encoded events with `Schema.UndefinedOr` (or loose `Schema.optional`) fields produce args of shape `{ ..., flag: undefined }`, while JSON wire transport drops the `undefined`-valued key, leaving `{ ... }`. The current `isEqualEncoded` comparator compares args with `deepEqual`, which counts keys via `Object.keys` and treats the two forms as unequal. The sync merge therefore falsely takes the rebase path when an event comes back from the sync provider, and state-dependent materializers re-run on already-mutated state, surfacing as `MaterializerHashMismatchError`.

## Reproduction

Community-provided repo with a Cloudflare Durable Object sync server and a sparse event commit: https://github.com/Fuzzyma/livestore-hash-missmatch (see `src/live-schema.ts` for the relevant `Schema.UndefinedOr` fields).

## Regression window

Surfaced after #1161 (`65ca2a1bd`) replaced `JSON.stringify(a.args) === JSON.stringify(b.args)` with `deepEqual(a.args, b.args)` to fix a key-order bug. The previous `JSON.stringify` form masked this issue because both sides were stringified through `JSON.stringify`, dropping `undefined` symmetrically. The structural `deepEqual` exposes the asymmetry.

## Root cause

`Schema.encodeUnknownSync` for a Struct with `Schema.UndefinedOr(T)` materializes omitted properties as `{ ..., key: undefined }`. JSON wire transport drops these, so the local pending event and the same event after a wire round-trip have different `Object.keys` lengths and compare unequal under `deepEqual`.

This is a JSON-wire/in-memory impedance mismatch: Effect Schema's encoded type system distinguishes "property present with value `undefined`" from "property absent", but JSON does not.

## Impact

Any sync flow where an event schema uses `Schema.UndefinedOr` or loose `Schema.optional` and the field is omitted at commit time. With state-dependent materializers, the resulting rebase rerun produces different SQL bindings than the first run, triggering `MaterializerHashMismatchError` in dev mode. Without dev hash checks, the user observes duplicate row insertion or other state divergence.

Issue #1131, opened 2026-03-25
Handle empty inserts for schema tables with only optional fields


IGassmann:
## Summary

This is related to #1117.

The typing fix for `table.insert()` now allows `insert({})` when a schema-derived table has no required insert columns, but the SQLite query builder still emits invalid SQL for that case.

## Reproduction

```ts
import { Schema as S, State } from "@livestore/livestore"

const UserSchema = S.Struct({
  bio: S.optional(S.String),
}).annotations({ title: "users" })

const users = State.SQLite.table({ schema: UserSchema })

users.insert({})
```

With the current builder this generates:

```sql
INSERT INTO "users" () VALUES ()
```

SQLite rejects that with a syntax error near `)`.

## Expected behavior

One of these needs to happen consistently:

- `insert({})` should be rejected for schema-derived tables whose insertable columns are all optional, or
- empty inserts should compile to valid SQLite, e.g. `INSERT INTO "users" DEFAULT VALUES`

## Why this matters

#1117 fixed the TypeScript side by making encoded-optional schema fields omittable in `table.insert()`. That exposed a runtime path where previously-invalid calls now typecheck but fail when executed.

## Notes

The review that surfaced this points at `packages/@livestore/common/src/schema/state/sqlite/table-def.ts`, where the new optionality typing now makes this empty-insert case reachable.

Issue #1117, opened 2026-03-22
Typing bug with `table.insert()`


Sam-Kruglov:
### Bug Description

When defining a SQLite table using an Effect Schema via State.SQLite.table({ schema }), the insert() method on the resulting table object treats all columns as required at the TypeScript level. Even if a field is marked as optional (S.optional) or nullable (S.NullOr) in the schema, TypeScript expects it to be present in the insert() call.
The issue seems to stem from how livestore derives the internal Columns type from an Effect Schema. The type-level mapping (specifically in SchemaToColumns.FromTypes) results in column definitions where the nullable property is typed as a general boolean. Because the internal type utility RequiredInsertColumns checks if TColumns[K]['nullable'] extends true, the boolean type fails this check, causing the field to be treated as required.
Note that update() is not affected because it correctly uses Partial<RowSchema>.

### Minimal Reproduction

```ts
import { Schema as S, State } from '@livestore/livestore';

// 1. Define a schema with optional/nullable fields
const UserSchema = S.Struct({
  id: S.String.pipe(State.SQLite.withPrimaryKey),
  name: S.String,
  // This field is optional and nullable
  bio: S.optional(S.NullOr(S.String)),
}).annotations({ title: 'users' });

// 2. Create the table
const usersTable = State.SQLite.table({
  schema: UserSchema,
});

// 3. Attempting to insert without the optional field
// @ts-expect-error - Currently fails with: 
// Property 'bio' is missing in type '{ id: string; name: string; }' 
// but required in type '{ readonly id: string; readonly name: string; readonly bio: string | null; }'.
usersTable.insert({
  id: 'user_1',
  name: 'Alice',
});
```

### LiveStore Version

0.4.0-dev.22

### Environment

```markdown
- OS: mac
- Node.js version: 
- Package manager: pnpm
- Runtime: (browser)
```

### Expected Behavior

Fields that are marked as S.optional or S.NullOr in the Effect Schema should be optional in the insert() method's input type.

### Actual Behavior

TypeScript requires all schema fields to be present in the insert() call, regardless of their optionality in the schema.

### Error Logs/Screenshots

```shell

```

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before
- [x] I have provided a minimal reproduction that demonstrates the issue
- [x] I have included all the requested information above

Issue #518, opened 2025-07-26
Add schema-first table definitions with @effect/sql Model integration


schickling:
## Summary

Add support for schema-first table definitions that can integrate with `@effect/sql` Model classes, providing a single source of truth for table schemas and enabling better integration with Effect's ecosystem.

## Background

Currently, LiveStore uses a column-first approach for table definitions:

```typescript
export const tables = {
  recipes: State.SQLite.table({
    name: "recipes",
    columns: {
      id: State.SQLite.text({ primaryKey: true }),
      title: State.SQLite.text({ default: "" }),
      imageUrl: State.SQLite.text({ nullable: true }),
      prepTime: State.SQLite.real({
        nullable: true,
        schema: Schema.DurationFromMillis,
      }),
      cookingTime: State.SQLite.real({
        nullable: true,
        schema: Schema.DurationFromMillis,
      }),
      rating: State.SQLite.real({ nullable: true }),
      servings: State.SQLite.real({ nullable: true }),
      ingredients: State.SQLite.json({
        schema: Schema.Array(IngredientsComponent),
      }),
      steps: State.SQLite.json({ schema: Schema.Array(Step) }),
      createdAt: State.SQLite.integer({
        schema: Schema.DateTimeUtcFromNumber,
        default: {
          sql: "(unixepoch('subsec') * 1000)",
        },
      }),
      updatedAt: State.SQLite.integer({
        schema: Schema.DateTimeUtcFromNumber,
        default: {
          sql: "(unixepoch('subsec') * 1000)",
        },
      }),
      deletedAt: State.SQLite.integer({
        nullable: true,
        schema: Schema.DateTimeUtcFromNumber,
      }),
    },
  }),
}
```

This approach requires defining both the table structure and any associated schemas separately, leading to potential inconsistencies and maintenance overhead.

## Proposed Solution

Add support for a `schema` property at the table level that accepts Effect Schema definitions, enabling a schema-first approach with `@effect/sql` Model integration:

```typescript
export class Recipe extends Model.Class<Recipe>("Recipe")({
  ...ExtractedRecipe.fields,
  id: Schema.String.annotations({
    description: "A unique identifier for the recipe.",
  }),
  cookingTime: Schema.NullOr(Schema.DurationFromMillis),
  prepTime: Schema.NullOr(Schema.DurationFromMillis),
  ingredients: Model.JsonFromString(Schema.Array(IngredientsComponent)),
[...]

PR #1308, opened 2026-06-09
refactor!: remove schema-derived `State.SQLite.table({ schema })`, make explicit columns the one way

### LINKED ISSUE #1307: Remove schema-derived `State.SQLite.table({ schema })`; make explicit columns the one way to define state tables
## Problem

`State.SQLite.table()` accepts two ways to define a table:

```ts
// 1. Explicit columns (every example uses this)
State.SQLite.table({
  name: 'users',
  columns: { id: State.SQLite.text({ primaryKey: true }), age: State.SQLite.integer({ nullable: true }) },
})

// 2. Derived from an Effect Schema
const User = Schema.Struct({ id: Schema.String.pipe(State.SQLite.withPrimaryKey), age: Schema.optional(Schema.Int) })
State.SQLite.table({ schema: User })
```

Approach 2 was added to provide a single source of truth for table schemas (#518, implemented in #544). In practice it forces a lossy mapping: Effect's type system is strictly richer than SQLite's storage model, so deriving columns from a schema must **collapse** distinct states and **guess** storage affinity.

**Collapse — `null` vs `undefined` vs absent.** SQLite has one notion of absence (`NULL`). The derivation folds optional / `null` / `undefined` into a single `nullable` flag:

```ts
// column-def.ts:30
const isNullable = forceNullable === true || hasNull(ast) === true || hasUndefined(ast) === true
```

So `Schema.optional(Schema.NullOr(Schema.String))` — three distinct TS intents ("not provided", "explicitly null", a string) — becomes one nullable `TEXT` column. After a write→read round-trip you only ever read back `null`; "not provided" and "explicitly null" are indistinguishable. The code knows this is lossy and only warns (`column-def.ts:78-82`).

**Guess — affinity inference.** `Schema.Number` carries no integer-vs-float information (JS has one number type), so the derivation guesses: it sniffs identifier annotations (`Int`, `DateFromNumber`) and otherwise defaults to `REAL` (`column-def.ts:174-181`). A field intended as integers silently becomes a `REAL` column unless the user opts in with `Schema.Int`.

This has produced a recurring stream of schema-table-specific bugs, each patching a symptom of the same mismatch rather than the root:

- #1117 → #1128: schema-derived `insert()` wrongly required nullable columns; fixed by making the **types agree with** the lossy collapse.
- #1131: that fix then exposed an invalid-SQL empty-insert path, reachable only for schema-derived tables.
- #1217: the same `undefined`-vs-absent impedance mismatch resurfacing in the sync layer (`Schema.UndefinedOr` / loose `Schema.optional`).

The explicit-columns approach has none of this: the user states the affinity (`integer` vs `real`) and nullability directly, `null` enters through exactly one explicit gate (`nullable: true` → `Schema.NullOr`, `field-defs.ts:107`), and there is nothing to infer or collapse.

## Proposal

Keep only the explicit-columns approach. Remove the schema-derived table definition:

- Drop `table()` overloads 2 & 3 and the `schema` branch of the implementation (`table-def.ts:148-181`, `:204+`).
- Remove the schema→column inference engine (`column-def.ts`).
- Remove the column-metadata annotation helpers that exist only to feed the schema path — `withPrimaryKey` / `withUnique` / `withAutoIncrement` / `withDefault` / `withColumnType` and their annotation symbols (`column-annotations.ts`). In the explicit path this metadata is passed as plain arguments (`{ primaryKey: true }`), so these become dead code.

Effect Schema is **not** being removed from columns — each explicit column still carries a per-value schema for encode/decode (`State.SQLite.text({ schema })`, `json`, `datetime`, `boolean`). Only "derive the whole table from one `Struct`" goes away.

`State.SQLite.clientDocument()` is **unaffected**: it accepts a value schema but stores the document as a single `json` column and builds its table via the column form (`client-document-def.ts:72-77`), not via the schema→column engine.

## Tradeoffs / open questions

- **Breaking change, but near-zero internal blast radius.** No example and no non-test package source uses the `schema:` form — the only in-repo users are test coverage and the `table-def.ts` JSDoc; system tables and `clientDocument` use the column form. External adoption is unknown (the API shipped in #544, Aug 2025). Decide between a hard cut and a deprecate-then-remove window, and provide a migration note (schema form → explicit columns).
- **Issue housekeeping.** This reverses #544 (which implemented #518). #470 (the separate, still-open schema-first proposal) should be closed as decided-against, and #1131 becomes moot once schema tables are gone.
- **Lost ergonomics.** The schema-first "single source of truth" goes away; explicit columns are more verbose. The position is that explicitness is worth it — it is predictable and never silently lossy.
- **Public surface.** `getColumnDefForSchema` is re-exported from the package entry (`table-def.ts:12`) with no internal consumers; removing it is part of the public API cut.
- **Contradicting TODO.** `table-def.ts:149` currently says "drop support for `column` when Effect Schema v4 is released" — the *opposite* direction. This proposal supersedes it; confirm before proceeding.
- **Effect v4 timing.** Worth deciding whether this lands before or after the v4 migration.

bohdanbirdie:
## Problem

`State.SQLite.table()` accepts two ways to define a table: explicit columns, and deriving the whole table from an Effect `Schema.Struct`. The schema-derived form is inherently **lossy** because SQLite's storage model is strictly poorer than Effect's type system:

- **Collapse — `null` vs `undefined` vs absent.** SQLite has one notion of absence (`NULL`). The derivation folds `optional` / `null` / `undefined` into a single `nullable` flag, so e.g. `Schema.optional(Schema.NullOr(Schema.String))` becomes one nullable `TEXT` column; after a write→read round-trip "not provided" and "explicitly null" are indistinguishable. The code knew this was lossy and only `console.warn`-ed.
- **Guess — affinity inference.** `Schema.Number` carries no integer-vs-float information, so the derivation guessed: sniff `Int` / `DateFromNumber` identifier annotations, otherwise default to `REAL`.

This produced a recurring stream of schema-table-only bugs, each patching a symptom of the same mismatch:

- #1117 → #1128: schema-derived `insert()` wrongly required nullable columns; fixed by making the **types agree with** the lossy collapse.
- #1131: that fix then exposed an invalid-SQL empty-insert path, reachable only for schema-derived tables.
- #1217: the same `undefined`-vs-absent mismatch resurfacing in the sync layer.

The design decision to remove the schema-derived form was discussed and approved in #1307.

## Solution

Make **explicit columns the one way** to define state tables. The user states affinity and nullability directly, `null` enters through exactly one gate (`nullable: true` → `Schema.NullOr`), and there is nothing to infer or collapse.

- Drop the `table()` schema overloads and the schema branch of the implementation.
- Remove the schema→column inference engine (`column-def.ts`).
- Remove the annotation helpers that only fed that path — `withPrimaryKey` / `withUnique` / `withAutoIncrement` / `withDefault` / `withColumnType` (and `getColumnDefForSchema`). In the explicit path this metadata is passed as plain arguments (`{ primaryKey: true }`), so these were dead weight.
- Update tests and docs; remove the "SQLite state schema (Effect schema)" page and its snippets.

**Effect Schema is _not_ removed from columns** — each explicit column still carries a per-value schema for encode/decode (`text({ schema })`, `json`, `datetime`, `boolean`). `State.SQLite.clientDocument()` is **unaffected** (it stores the document as a single `json` column via the column form, not via the schema→column engine).

**Trade-off:** the schema-first "single source of truth" ergonomics go away and explicit columns are more verbose. The position (per #1307) is that predictable, never-silently-lossy columns are worth it.

**Migration:** replace `table({ schema: MySchema })` with `table({ name, columns })`, declaring each column's affinity and nullability directly.

## Validation

- `vitest run src/schema/` in `@livestore/common` → **111 passed**.
- `tsc --build packages/@livestore/common` → **exit 0**.
- Repo-wide sweep: no remaining references to the removed API (engine, helpers, dead types, table-level `schema:`); column-level `schema:` and `clientDocument({ schema })` are retained.
- Independent code review (re-ran the tests + typecheck and the dangling-reference sweep) → clean.
- ⚠️ **Not yet run locally:** `dt lint:full:fix` (oxlint/oxfmt) and full-repo `dt ts:check` — relying on CI / a follow-up local pass to confirm. (Marking this PR a draft for that reason.)

Changeset added: `@livestore/common: minor` (pre-1.0 breaking; packages are in a `fixed` group).

## Related issues

- Closes #1307
- Reverses #544 (which implemented #518)
- #470 (separate, still-open schema-first proposal) should be closed as decided-against
- #1131 becomes moot once schema-derived tables are gone

Suggested labels (I lack triage permission): `design decision`, `schema`, `effect`, `breaking-change`.


IGassmann (approved):

acusti:
i get this logic, but i don't like the end result. i migrated to schema-first when #544 was merged and i like it better. i understand from the PR description that this had already been deliberated and decided upon, so i'm not asking you all to reconsider it. just figured i should submit my opinion for the record.

IGassmann:
Hey @acusti ! Thanks for sharing your opinion. This is helpful.

Could you please share more about why you prefer the Effect Schema approach? We might at some point switch to Effect Schema (after Effect v4 migration) as being the default way, so this would be valuable information.

The main goal for us is to avoid supporting two approaches, and currently the `table({ name, columns })` approach is the most functional.

packages/@livestore/common/src/schema/state/sqlite/table-def.ts
```diff
@@ -95,59 +86,28 @@ export type TableOptions = {
...
  * // Adding indexes
- * const PostSchema = Schema.Struct({
- *   id: Schema.String.pipe(State.SQLite.withPrimaryKey),
- *   title: Schema.String,
- *   authorId: Schema.String,
- *   createdAt: Schema.Date,
- * }).annotations({ identifier: 'posts' })
- *
  * const postsTable = State.SQLite.table({
- *   schema: PostSchema,
```
bohdanbirdie, line 133: - effect 4 schema might be more capable
- runtime matters more than sqlite, the runtime is practically a source truth here
- bijective???

1. is schema-first approach feasible?
2. keep this PR open, create a parallel PR that explores effect v4 schema as alternative and a single unified approach

Issue #1328, opened 2026-06-12
Sync backend Durable Object never hibernates — Effect.never's setInterval keepalive blocks CF WebSocket Hibernation


bohdanbirdie:
## Summary
Every live WebSocket sync connection to a `@livestore/sync-cf` Durable Object holds pending `setInterval(…, 2**31-1)` timers for the connection's entire lifetime. A pending timer is an explicit disqualifier for Cloudflare DO **WebSocket Hibernation**, so the sync DO is kept resident in memory (and billed for wall-clock duration) instead of hibernating between messages. Root cause: `effect`'s `Effect.never` is implemented with a real `setInterval` (a Node event-loop keepalive), and the WS-RPC server parks three connection-holding fibers on it.

## Impact
- Durable Objects bill for **wall-clock residency**, not CPU. A DO that can't hibernate is billed for the full duration of every open connection, even while completely idle.
- A sync client holds a **long-lived, mostly-idle** connection by design (to receive live updates) — so the worst case *is* the normal operating mode.
- Sites 1️⃣ and 2️⃣ have existed since LiveStore's first Cloudflare adapter, so **every** CF-DO sync deployment ships a backend that cannot hibernate. It's latent at low connection volume (free-tier duration absorbs it) and becomes an escalating cost as concurrent/long-lived connections grow.

## Root cause
`effect`'s `never` parks a fiber forever using a `setInterval`:

```js
// effect/dist/esm/internal/core.js
export const never = asyncInterrupt(() => {
  const interval = setInterval(() => {}, 2 ** 31 - 1) // ≈ 24.8 days, no-op callback
  return sync(() => clearInterval(interval))
})
```

The fiber never-resolves because `asyncInterrupt` never calls `resume`; the `setInterval` is a separate concern — a **Node** event-loop keepalive that stops the *process* from exiting when nothing else is pending. In a Worker/DO that keepalive is (a) unnecessary (the platform keeps the isolate alive while the socket is open) and (b) actively harmful: it is exactly the "pending work" that suppresses hibernation. Per Cloudflare's DO lifecycle docs, one hibernation gate is verbatim:

> "No `setTimeout`/`setInterval` scheduled callbacks are set, since there would be no way to recreate the callback after hibernating."

`Stream.never` and `Layer.launch` both bottom out in `Effect.never`, so they carry the same timer.

## Where it bites (the three parks)
[...]

PR #1338, opened 2026-06-23
DO hibernation, end to end: idle sync-DO hibernation + DO-RPC live-pull survival

bohdanbirdie:
> The #1328 hibernation chain — four bugs in two symmetric server/client pairs, shipped as two commits. "Bug N" is internal chain numbering (not GitHub issue references).

## Problem

Cloudflare bills Durable Objects for **wall-clock residency**, so an idle DO is **hibernated** (evicted) and reconstructed on the next request. The asymmetry that drives everything: **persisted SQLite survives; in-memory JS — instance fields, module-level `Map`s — is wiped.**

The chain is four bugs forming **two symmetric pairs.** In each pair a DO *couldn't hibernate*, and *fixing that* exposed a second bug that only manifests once the DO actually evicts:

**Server side**
- **Bug 1 — the sync DO couldn't hibernate.** Effect's `Effect.never` parks a fiber by registering `setInterval(() => {}, 2**31-1)` — a Node-only event-loop keepalive that is useless in a Worker and is the *exact* CF hibernation disqualifier (an isolate with any pending `setInterval`/`setTimeout` cannot evict). The WS-RPC sync server held **three** such parks per idle live connection (`Layer.launch`, `@effect/rpc`'s `withRun` run-loop, and each live-pull `Stream.never`).
- **Bug 2 — once it *can* hibernate, it silently forgets its DO-RPC subscribers.** Live updates are fanned out by iterating a subscriber registry and calling each reader's `clientDo.syncUpdateRpc(chunk)` (a DO→DO **reverse RPC**). That registry was an **in-memory `Map`** — exactly the state hibernation wipes. The reconstructed sync DO loops over zero subscribers, and a reader that subscribed earlier **stops receiving live updates with no error.** Fixing Bug 1 is what *unmasked* Bug 2.

**Client side (the mirror)**
- **Bug 3 — the client-host DO couldn't hibernate.** The same `Effect.never` root, on the DO-RPC client transport's run-loop.
- **Bug 4 — once it *can* hibernate, it can't receive after its own hibernation.** A reverse-RPC arriving while the reader DO is hibernated hits a DO with no live store to deliver into. Fixing Bug 3 unmasks Bug 4 (just as Bug 1 unmasks Bug 2).

**Commit split:** Bug 1 shipped first as **Commit 1** (the server timer-park); Bugs 2, 3, and 4 are **Commit 2**.

**Why this is structurally hard.** The obvious fix to Bug 2 — persist the registry in SQLite — fixes it but **deletes the only reaper**: the in-memory map's loss-on-eviction was the system's sole (accidental) garbage collection. A persisted registry never forgets, so dead subscriptions accumulate forever (unbounded SQLite growth, O(N) fan-out, and the cost of waking dead DOs on every push). And the sync DO **structurally cannot detect a dead subscriber**:

- a reverse-RPC to a hibernated/retired client **resurrects it and succeeds** — the probe revives the corpse;
- there is **no destructor hook** on hibernation/retirement; and
- the DO-RPC pull is an **ending stream** — it drains the backlog and completes, so (unlike a WebSocket, whose *disconnect* the server observes) there is no close signal.

So *reap-on-failure* is both trigger-happy (a transient hiccup false-reaps a **live** client → silent permanent severance) and blind (a genuinely **dead** client never produces a failure). The failure channel is unusable as a reaping signal — **only the client knows whether it still wants the subscription.**

**Constraints (hard).** No new public API on `@livestore` packages; no user-facing config knobs; **no time-based mechanisms** (alarm / lease / TTL); zero regressions (correctness, billing, storage).

### Commit 1 (Bug 1) — `47858396f`

Replace each `Effect.never` park with a timer-less `Effect.async<never>(() => {})` (identical never-resolving, interruptible behavior, no timer): in `common-cf` ws-rpc-server (built into the connection scope with a hand-rolled protocol `run` loop instead of `@effect/rpc`'s `withRun`) and the `sync-cf` live-pull handler. Idle ⇒ 0 long-period timers ⇒ hibernation-eligible.

### Commit 2 (Bugs 2 / 3 / 4) — reap on the client's own verdict

Since the producer can't detect liveness, the **client reports it** through the reverse-RPC's return value, and the sync DO reaps on that:

```ts
async syncUpdateRpc(payload): Promise<SyncUpdateRpcResult> {
  if (this.store === undefined) return true   // store gone (hibernated) → ask to be reaped
  return handleSyncUpdateRpc(payload)         // store live → route the chunk, return void
}
```

- **Persist the registry** in SQLite (one `rpc_subscription` row per subscribing caller DO) so it survives hibernation — fixes Bug 2.
- **Reap only on the explicit `true` verdict.** Failure/timeout is **transient — keep the row** (a transient fault against a live client must never reap it). The broadcast loop reads the verdict and `DELETE`s the row only on `true`.
- **The wake and the reap are the same event.** The first push after a client hibernates wakes it once; its reconstructed (storeless) gate returns `true`; the row is reaped; it sleeps again. The registry **self-cleans** instead of waking dead DOs forever — each hibernated subscriber is woken at most once per hibernation cycle.
- **Bug 4 dissolves** into that one-line gate — a woken storeless client returns `true` and recovers via catch-up pull on next use. No boot-on-wake; the old keep-alive `alarm()` / wake-resubscribe machinery is **removed** from the examples.
- **Bug 3** parks the client-host DO-RPC transport's run-loop on the same timer-less `Effect.async`, so a DO-resident client can hibernate too (the client-side mirror of Bug 1).

**Two correctness guards:**

- **Per-call timeout** (`Effect.disconnect` + `Effect.timeout`, 5s). The broadcast loop is `Effect.uninterruptible` (so it can't be killed mid-fan-out); without a per-call bound, one hung `syncUpdateRpc` is in-flight I/O that pins the sync DO resident and **silently re-breaks Bug 1**. `timeout` bounds each call; `disconnect` makes the *call* interruptible inside the uninterruptible loop so the timeout actually releases the CF outbound I/O.
- **Monotonic `generation` compare-and-delete token.** The loop reads a snapshot and can run for seconds; meanwhile the same DO may re-subscribe (`INSERT OR REPLACE`, new generation). `remove(doId, generation)` deletes only if the generation still matches, so a stale reap cannot clobber a fresh re-subscribe. It is a counter seeded from `MAX(generation)` (to stay increasing across hibernation), **not** `Date.now()` — CF freezes the clock within a turn, so same-turn re-subscribes would otherwise collide.

**Why store-liveness, not mailbox-presence (the one real design fork).** Gating on `cachedStore === undefined` rather than "is there a routing mailbox" deliberately dissolves the **subscribe boot-race**: the server writes the subscription synchronously on `Pull`, but the client registers its mailbox only *after* draining the backlog. A push in that window finds no mailbox — under a mailbox-presence probe that would false-reap an about-to-be-live subscriber and permanently sever it. Under store-liveness the store is live throughout the window, so the chunk is dropped as transient and recovered by the finishing pull. The library's `handleSyncUpdateRpc` is therefore **routing-only and never reaps**; the consumer gate is the *only* reap signal.

**Client-side hygiene (the "keystone").** The reverse-RPC routing map (`requestIdMailboxMap`) now drops its entry on scope close. Without it, a closed/reconnected pull leaves a shut-down mailbox in the module-level map — and `offer` on a shut-down mailbox is a silent no-op, so a stale entry is an indistinguishable false-alive plus a leak across store boots in a warm isolate. (Defense-in-depth, *not* the reap signal.) A single module map routes correctly across multiple stores because `@effect/rpc` request IDs are isolate-unique — no per-`storeId` scoping needed.

**Consumer cooperation — and the one gotcha.** The gate is ~2 lines reading a store-liveness field every consumer already maintains; the return-type widening (`Promise<void>` → `Promise<void | boolean>`) is source-compatible, so existing `return handleSyncUpdateRpc(payload)` consumers still type-check. **Clear-on-shutdown is load-bearing:** hibernation wipes the liveness flag for free, but a *graceful* close must reset it or the gate never fires. (Will be documented for adopters in the docs follow-up.)

**Behavioral contract — "live only while the store is alive, ≈ WebSocket."** Idle ⇒ reaped; recovery is a normal store boot + catch-up pull from the cursor on next use (the *same* code path a WS reconnect uses → no data loss). Nothing wakes a hibernated client to keep it synced — an always-on reactor that hibernates *and* wakes on every event is a contradiction (keep the store alive if you need that).

**Rejected alternatives.** *Lease/alarm/TTL reaper* — needs time (ruled out) and a client heartbeat that itself defeats hibernation; false-reaps an idle-but-alive subscriber. *Persisted "retired" marker* — can't catch vanished clients (no destructor). *Adapter-owned base class injecting `syncUpdateRpc`* — the adapter has no per-instance store handle, so it would require a class users `extends` = new public API (forbidden), for identical correctness.

**Trade-off / known follow-up (separate PR).** Live delivery is best-effort. A transient drop to a *still-resident* client leaves a forward gap the sync core currently advances past silently (no `parentSeqNum` contiguity check on the upstream boundary). This is a **pre-existing** weakness, first *exposed* by DO-RPC keep-on-failure; the proper fix (sync-core gap-detection → re-pull from cursor) is transport-agnostic and tracked for a follow-up PR rather than bolted onto this one.

### Implementation map

| Concern | Files |
| --- | --- |
| Bug 1 — server timer-park (Commit 1) | `common-cf/.../ws-rpc/ws-rpc-server.ts`, `sync-cf/.../do/*` live-pull handler |
| Bug 3 — client timer-park | `common-cf/src/do-rpc/client.ts` (`makeProtocol`) |
| Verdict type + producer returns it | `common-cf/src/do-rpc/server.ts` (`SyncUpdateRpcResult`, `emitStreamResponse`) |
| Persisted registry + `generation` | `sync-cf/.../do/sqlite.ts`, `cf-worker/shared.ts`, `do/layer.ts`, `do/transport/do-rpc-server.ts` |
| Reap loop + two guards | `sync-cf/.../do/push.ts` |
| Keystone + routing-only handler | `sync-cf/src/client/transport/do-rpc-client.ts` |
| Consumer gate (Bug 4 dissolves) | integration fixture + `cloudflare-todomvc`, `web-email-client/{Mailbox,Thread}ClientDO` |

## Validation

Host type-check (`tsc --build`) is clean for all changed files (2 unrelated pre-existing fixture errors remain). The tests lean on four rigor techniques — **prove the precondition** (a per-instance UUID that changes only on real reconstruction, so a green test can't be a false positive where nothing evicted), **isolate the variable** (a non-live pull confirms the server stored the event, so a missing live delivery means delivery failed, not data lost), **observe without perturbing** (a non-booting eventlog probe, since booting would trigger the very catch-up we're testing for the *absence* of), and **deterministic trigger where reproduction is unreliable** (a store-liveness switch that fires the exact gate when real hibernation can't be forced):

- **sync-provider — reap mechanism:** register a live subscription → flip the store-liveness switch → push → assert the persisted subscription count goes **1 → 0** → the closed client misses the live event → a catch-up pull recovers it.
- **sync-provider — Bug 2 regression under real sync-DO eviction:** idle the sync DO out (proven evicted via the UUID), assert the resident client still receives live (persisted registry survived), with a non-live pull control.
- **integration (workerd) — inverted contract under real client-host eviction:** a post-hibernation event is **reaped, not delivered live** (non-booting eventlog probe stays flat), then a control boot catch-up-pulls it.
- **integration — timer guard (Bug 3):** a DO-resident client holds **0 long-period timers** (can hibernate); a re-introduced `Effect.never` would fail it.

**Local vs. deployed scope note.** DO *instance* eviction is reproduced and trusted locally (instance fields are wiped on eviction, proven by the UUID), so the reap is exercised under real local eviction. The only residual is real-CF *isolate* teardown (module globals), which local dev keeps warm — and the design was deliberately built to not depend on it.

Still pending before this draft is merge-ready: `CHANGELOG` entry, docs snippet + adapter `.mdx`, and the full devenv `lint`/`ts` suites.

### Demo

Verdict round-trip — the reap mechanism:

```
push.ts broadcast (per subscriber, per chunk)         consumer DO
  rpcSubscriptions.all()  ← persisted rows (survive hibernation)
  emitStreamResponse(…) ──reverse RPC──▶ syncUpdateRpc(payload)
    .pipe(disconnect, timeout 5s, exit)   ◀── verdict ── (true | void)
      ├─ Exit.isFailure  → keep   (transient: never reap on failure)
      ├─ value === true  → reap this row (DELETE matched on `generation`)
      └─ value void      → next chunk (alive)
```

The two symmetric pairs:

```
            cannot hibernate         →  fixing it unmasks
server      Bug 1 (sync-DO park)     →  Bug 2 (forgets DO-RPC subscribers)
client      Bug 3 (client-DO park)   →  Bug 4 (can't receive after own hibernation)
```

## Related issues

- livestorejs/livestore#1328

### DIFF
--- a/packages/@livestore/sync-cf/src/client/transport/do-rpc-client.ts
+++ b/packages/@livestore/sync-cf/src/client/transport/do-rpc-client.ts
@@ -24,7 +24,7 @@ import type { SyncMetadata } from '../../common/sync-message-types.ts'
 
 export interface SyncBackendRpcStub extends CfTypes.DurableObjectStub, SyncBackendRpcInterface {}
 
-// TODO we probably need better scoping for the requestIdMailboxMap (i.e. support multiple stores, ...)
+// requestIds are isolate-unique (`@effect/rpc` module-level counter), so one map routes correctly across stores.
 type EffectRpcRequestId = string // 0, 1, 2, ...
 const requestIdMailboxMap = new Map<EffectRpcRequestId, Mailbox.Mailbox<SyncMessage.PullResponse>>()
 
@@ -82,11 +82,16 @@ export const makeDoRpcSync =
                   if (res._tag === 'None')
                     return shouldNeverHappen('There should at least be a no-more page info response')
 
+                  const rpcRequestId = res.value.rpcRequestId
+
                   const mailbox = yield* Mailbox.make<SyncMessage.PullResponse>().pipe(
-                    Effect.acquireRelease((mailbox) => mailbox.shutdown),
+                    // On scope close, drop the map entry so a closed client leaves no stale route.
+                    Effect.acquireRelease((mailbox) =>
+                      mailbox.shutdown.pipe(Effect.zipRight(Effect.sync(() => requestIdMailboxMap.delete(rpcRequestId)))),
+                    ),
                   )
 
-                  requestIdMailboxMap.set(res.value.rpcRequestId, mailbox)
+                  requestIdMailboxMap.set(rpcRequestId, mailbox)
 
                   return Mailbox.toStream(mailbox)
                 }).pipe(Stream.unwrapScoped),
@@ -159,17 +164,14 @@ export const makeDoRpcSync =
     }).pipe(Effect.withSpan('rpc-sync-client:makeDoRpcSync'))
 
 /**
+ * Routes a reverse-RPC chunk into the matching live pull stream. A client DO implementing `syncUpdateRpc`
+ * must gate on its own store liveness — return `true` when the store is gone so the sync DO reaps the
+ * subscription (recovery is a normal catch-up pull on next use, ≈ WebSocket reconnect).
  *
  * ```ts
- * import { DurableObject } from 'cloudflare:workers'
- * import { ClientDoWithRpcCallback } from '@livestore/common-cf'
- *
- * export class MyDurableObject extends DurableObject implements ClientDoWithRpcCallback {
- *   // ...
- *
- *   async syncUpdateRpc(payload: RpcMessage.ResponseChunkEncoded) {
- *     return handleSyncUpdateRpc(payload)
- *   }
+ * async syncUpdateRpc(payload: RpcMessage.ResponseChunkEncoded): Promise<SyncUpdateRpcResult> {
+ *   if (this.#store === undefined) return true
+ *   return handleSyncUpdateRpc(payload)
  * }
  * ```
  */
@@ -181,10 +183,9 @@ export const handleSyncUpdateRpc = (payload: unknown) =>
     const pullStreamMailbox = requestIdMailboxMap.get(decodedPayload.requestId)
 
     if (pullStreamMailbox === undefined) {
-      // Case: DO was hibernated, so we need to manually update the store
-      yield* Effect.log(`No mailbox found for ${decodedPayload.requestId}`)
+      // Transient: live pull not (re)registered yet. Drop the chunk; the next pull catches up from the cursor.
+      yield* Effect.logDebug(`No mailbox for requestId ${decodedPayload.requestId}; dropping chunk (transient)`)
     } else {
-      // Case: DO was still alive, so the existing `pull` will pick up the new events
       yield* pullStreamMailbox.offer(decoded)
     }
   }).pipe(Effect.withSpan('rpc-sync-client:rpcCallback'), Effect.tapCauseLogPretty, Effect.runPromise)
--- a/packages/@livestore/sync-cf/src/client/transport/do-rpc-client.ts
+++ b/packages/@livestore/sync-cf/src/client/transport/do-rpc-client.ts
@@ -87,7 +87,9 @@ export const makeDoRpcSync =
                   const mailbox = yield* Mailbox.make<SyncMessage.PullResponse>().pipe(
                     // On scope close, drop the map entry so a closed client leaves no stale route.
                     Effect.acquireRelease((mailbox) =>
-                      mailbox.shutdown.pipe(Effect.zipRight(Effect.sync(() => requestIdMailboxMap.delete(rpcRequestId)))),
+                      mailbox.shutdown.pipe(
+                        Effect.zipRight(Effect.sync(() => requestIdMailboxMap.delete(rpcRequestId))),
+                      ),
                     ),
                   )
 
--- a/examples/cloudflare-todomvc/src/live-store-client-do.ts
+++ b/examples/cloudflare-todomvc/src/live-store-client-do.ts
@@ -1,5 +1,4 @@
 import { DurableObject } from 'cloudflare:workers'
-import type { AlarmInvocationInfo } from '@cloudflare/workers-types'
 
 import { type ClientDoWithRpcCallback, createStoreDoPromise } from '@livestore/adapter-cloudflare'
 import { nanoid } from '@livestore/livestore'
@@ -65,15 +64,11 @@ export class LiveStoreClientDO extends DurableObject<Env> implements ClientDoWit
         console.log(`todos for store (${this.storeId})`, todos)
       })
     }
-
-    await this.ctx.storage.setAlarm(Date.now() + 1000)
-  }
-
-  alarm(_alarmInfo?: AlarmInvocationInfo): void | Promise<void> {
-    this.subscribeToStore()
   }
 
   async syncUpdateRpc(payload: unknown) {
-    await handleSyncUpdateRpc(payload)
+    // Store gone (hibernated): ask the sync DO to drop this subscription; it recovers on next use.
+    if (this.hasCachedStore === false) return true
+    return handleSyncUpdateRpc(payload)
   }
 }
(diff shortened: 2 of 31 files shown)


packages/@livestore/sync-cf/src/client/transport/do-rpc-client.ts
```diff
@@ -24,7 +24,7 @@ import type { SyncMetadata } from '../../common/sync-message-types.ts'
 
 export interface SyncBackendRpcStub extends CfTypes.DurableObjectStub, SyncBackendRpcInterface {}
 
-// TODO we probably need better scoping for the requestIdMailboxMap (i.e. support multiple stores, ...)
+// requestIds are isolate-unique (`@effect/rpc` module-level counter), so one map routes correctly across stores.
```
bohdanbirdie, line 27: Removed the prior `// TODO: better scoping for requestIdMailboxMap (support multiple stores)` that lived here — resolved, not just dropped. `@effect/rpc` assigns request IDs from a module-level, isolate-unique counter, so a single module-level `requestIdMailboxMap` routes reverse-RPC chunks correctly even with multiple stores in one host DO; no per-storeId scoping is needed. The keystone finalizer (delete-on-scope-close, just below) keeps that map from leaking stale entries across store boots.

bohdanbirdie:
This PR is outdated and the work will be followed in other PRs:
https://github.com/livestorejs/livestore/pull/1541

PR #1364, opened 2026-07-04
Explore Userland Initialization Patterns for Client-Only Events

TheAndrewJackson:
## Problem

The first-party ClientDocument API does a lot automatically: `State.SQLite.clientDocument(...)` creates a special table shape, derives a client-only `set` event and materializer, and `.get(...)` can seed a missing row by committing that generated `set` event before the query runs. That convenience is useful, but it hides important behavior behind the read path: a read can perform a write, the initialization point is implicit in the app flow, and the initial value comes from either `.get(..., { default })` or the table’s configured default. This example makes those pieces explicit.

## Solution

Model UI state with normal SQLite tables, explicit `Events.clientOnly(...)` events, and manual materializers instead of the higher-level ClientDocument abstraction.

Initialization becomes a deliberate userland step: read the target row, commit an explicit client-only event if it is missing, and then let descendants read the materialized row. This keeps the important pieces visible in application code: the SQLite table being initialized, the initial row value, the client-only event being committed, and the point in the app flow where that commit happens.

Example render-time initialization:
```tsx
function RenderEnsurePage() {
  const { storeOptions } = Route.useRouteContext()
  const store = useStore(storeOptions)

  console.log('hello')
  useEnsureClientOnlyRow({
    store,
    table: tables.threadListUi,
    id: rowId,
    defaultValue: { selectedThreadId: 'support-004', sortBy: 'receivedAt', sortDirection: 'desc' } as const,
    label: 'hook:thread-list-ui',
    event: ({ id, defaultValue }) => events.threadListUiEnsured({ id, ...defaultValue }),
  })
  console.log('world')

  return (
    <DemoFrame title="useEnsureClientOnlyRow">
      <section className="pattern-note">
        <p>The hook commits an explicit client-only ensure event during render, then descendants read the row.</p>
        <ClientOnlyDataSummary pattern="render hook" rowId={rowId} mailboxId={mailboxId} />
      </section>
      <ThreadList store={store} rowId={rowId} mailboxId={mailboxId} />
    </DemoFrame>
  )
}
```

## Derived Data

Rows that depend on derived async data are supported with an `enabled` flag on `useEnsureClientOnlyRow`. When `enabled` is `false`, the hook skips the ensure step entirely, so it does not read-and-create the row or commit the default client-only event. Once the source data is ready, `enabled` can flip to `true`, allowing the hook to derive the initial value from complete local data and commit the explicit initialization event.

An example of this pattern is in `examples/client-only-events-init-apis/src/routes/derived/default-with-readiness-marker.tsx`.

## Demo

Run the example with: `pnpm --filter livestore-example-client-only-events-init-apis dev`

image


examples/client-only-events-init-apis/src/routes/derived/default-with-readiness-marker.tsx
```diff
@@ -0,0 +1,119 @@
...
+    {
+      store,
+      table: tables.threadListUi,
+      id: rowId,
+      defaultValue: defaultThreadListUi,
+      label: sourceRowsAreReady === false ? `derived-waiting:${demoKey}` : `derived-ready:${demoKey}`,
+      event: ({ id, defaultValue }) => events.threadListUiEnsured({ id, ...defaultValue }),
+    },
+    { enabled: sourceRowsAreReady },
+  )
```
TheAndrewJackson, line 54: Overall - decent

notes
- YEET the current hook/api entirely

investigate
- look into having a "event querying api" in order to decouple from the read-model
  - need to index the event `name` column
  - number hash of the `name` and `args` with index
  - get performance characteristics of inserting many event with and without indexes

explore
```tsx
store.useEnsureEventCommited({event})
```

TheAndrewJackson: Investigate removing the need for the `enabled` api by restructing the UI itself and removing the problem

Issue #1356, opened 2026-06-30
Audit Effect v4 fork options across runtime and tests


IGassmann:
## Problem

During the Effect v4 migration, many runtime and test call sites were migrated with explicit fork options to preserve Effect v3 behavior, usually with this note:

> TODO: These options were set to preserve Effect v3 fork behavior while migrating to Effect v4. Verify if they're the most appropriate configuration for this specific fork.

The stack can merge with these compatibility choices, but they should not remain as unexplained migration debt. Some forks may need `startImmediately: true`, some may need `uninterruptible: 'inherit'`, some may need both, and some may be able to use Effect v4 defaults.

Representative call sites include:

* `packages/@livestore/utils/src/effect/Effect.ts`
* `packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts`
* `packages/@livestore/common/src/leader-thread/make-leader-thread-layer.ts`
* `packages/@livestore/livestore/src/store/store.ts`
* `packages/@livestore/utils/src/effect/WebChannel/WebChannel.ts`
* `packages/@livestore/webmesh/src/channel/*`
* integration and package-common tests

## Scope

* Find every `TODO: These options were set to preserve Effect v3 fork behavior...` occurrence.
* Review whether each fork actually needs `startImmediately`, `uninterruptible: 'inherit'`, both, or neither under Effect v4.
* Replace repeated inline policy with a named helper only if the same semantic policy is intentional across call sites.
* Remove the migration TODOs once each call site is justified.

## Acceptance criteria

* No remaining `preserve Effect v3 fork behavior` TODOs.
* Any retained non-default fork behavior has a short local explanation of why it is required.
* Relevant sync/store, WebChannel, webmesh, and integration tests pass.

## Context

Follow-up from the Effect v4 migration finalization work tracked by livestorejs/livestore#1321 and parent issue livestorejs/livestore#1310.

IGassmann:
I checked the v3 vs v4 fork startup behavior directly with small probes against local checkouts. The important finding: the default startup behavior did **not** change. In both Effect v3 and Effect v4, the forked child does not run before the parent continues past the fork call; it runs after yielding/scheduler progress.

What changed in v4 is that the fork APIs now expose `{ startImmediately: true }` as an explicit opt-in for inline startup. In the probe, v4 defaults produced `after fork childRan=false`, while `{ startImmediately: true }` produced `after fork childRan=true`.

[...]

Issue #1321, opened 2026-06-11
Effect v4: Changeset and final validation


IGassmann:
## Parent

#1310

## What to build

Add the breaking Changeset and restore the final validation gate for the Effect v4 migration stack.

This PR should make the stack ready to merge as one unit after package, docs, examples, and test workspace migration slices are complete.

## Scope

- Add a Changeset for the breaking Effect v4 migration.
- State that LiveStore now targets Effect v4.
- Mention public peer dependency changes.
- Mention any changed `@livestore/utils/effect` exports that consumers may notice.
- Run and fix final validation failures from the full stack.
- Capture intentionally postponed cleanup as follow-up issues where appropriate.

## Acceptance criteria

- [x] A Changeset marks the Effect v4 migration as breaking.
- [x] Peer dependency changes are documented.
- [x] Public facade/export changes are documented where relevant.
- [x] `dt ts:check` passes.
- [x] `dt lint:full` passes.
- [x] Relevant tests pass for affected packages and behavior.
- [x] Deferred cleanup is captured as follow-up work if discovered.
- [x] Final PR body states problem, solution, validation commands, breaking changes, and known trade-offs/follow-ups.

## Validation

- [x] `dt ts:check`
- [x] `dt lint:full`
- [x] Relevant tests for migrated behavior, including focused tests for Effect utility wrappers/facades, service/context helpers, schema encode/decode paths, sync/store behavior, and affected package/runtime behavior.

## Blocked by

- #1320

PR #1387, opened 2026-07-10
fix(sync-cf): repair Effect v4 Durable Object RPC boundaries

IGassmann:
## Problem

After the Effect v4 migration, the two Cloudflare Durable Object RPC providers each passed only 3 of the 16 shared sync-provider cases, while Mock, HTTP RPC, and WebSocket providers still passed all 16. The 13 failures split across three transport/lifecycle boundaries:

| Cases | Root cause |
| ---: | --- |
| 9 | The custom native DO-RPC server decoded only the outer MessagePack envelope and sent JSON-codec payloads directly to handlers. An omitted optional `rpcContext` therefore arrived as `null` and was dereferenced. |
| 2 | Live `PullResponse` callbacks sent Effect v4 runtime `Option` objects through Cloudflare native RPC, which rejected them with `DataCloneError`. |
| 2 | The test proxy issued another request before its long-lived `IsConnected` RPC stream was registered, producing interrupted-fiber timeouts under the v4 dispatcher. |

Fixing the startup ordering also exposed a latent contract mismatch: the native client began permanently connected and implemented `connect` as a no-op. Separately, the provider labelled D1 was actually exercising DO SQLite because the test worker never selected D1 storage.

## Solution

- Decode each native DO-RPC request payload once with `Schema.toCodecJson(rpc.payloadSchema)` and the registered handler context before dispatching unary or streaming handlers. Malformed payloads now return a schema-encoded defect exit.
- Encode and decode out-of-band live `PullResponse` values with the same JSON codec, producing clone-safe native-RPC values.
- Model the native client lifecycle as `false -> successful Ping -> true`, with `connect` performing that real ping.
- Start only the proxy's `IsConnected` registration immediately, before `GetMetadata` uses the same client.
- Split the test worker into explicit D1 and DO SQLite Durable Object classes and preserve Wrangler's existing class migration history.
- Add unary, streaming, and malformed-payload regressions at the custom protocol boundary.

The eager fork is deliberately local. Bare `forkScoped` was scheduled/deferred in both Effect v3 and v4, so `startImmediately: true` is not presented as v3 compatibility. It is a stronger ordering requirement for this specific RPC registration. `uninterruptible: "inherit"` was tested independently and did not fix the failures, so the generic `SubscriptionRef.fromStream` behavior remains unchanged.

## Validation

- `devenv shell -- dt lint:full:fix`
- `devenv shell -- dt ts:check`
- `devenv shell -- mono test unit` (all configured unit projects passed; common-CF: 24 passed, 1 skipped)
- `pnpm vitest run src/do-rpc/client.test.ts` (2 passed)
- `pnpm vitest run src/do-rpc/rpc.test.ts` (11 passed, 1 skipped)
- `pnpm vitest run src/sync-provider.test.ts` (112 passed)
- Focused native DO-RPC matrix after the final fork localization (32 passed)
- Fork control experiment across D1 and DO: `startImmediately: true` passed 4/4 affected invocations; bare `forkScoped` and `uninterruptible: "inherit"` each passed 0/4.

`dt test:run`, referenced by the contributor guide, is not defined in the current task graph; `mono test unit` was used for the repository-wide unit gate instead.

### Demo

| Provider | Before | After |
| --- | ---: | ---: |
| Mock | 16/16 | 16/16 |
| Cloudflare HTTP RPC (D1 / DO) | 16/16 | 16/16 |
| Cloudflare WebSocket (D1 / DO) | 16/16 | 16/16 |
| Cloudflare Durable Object RPC (D1) | 3/16 | 16/16 |
| Cloudflare Durable Object RPC (DO) | 3/16 | 16/16 |

## Related issues

- Follow-up to #1369
- Related to #1356
- Related to #1321

### DIFF
--- a/.changeset/fix-do-rpc-effect-v4-boundaries.md
+++ b/.changeset/fix-do-rpc-effect-v4-boundaries.md
@@ -1,0 +1,6 @@
+---
+"@livestore/common-cf": patch
+"@livestore/sync-cf": patch
+---
+
+Fix Cloudflare Durable Object RPC sync providers failing to pull or receive live events after the Effect v4 migration.
--- a/packages/@livestore/common-cf/src/do-rpc/client.test.ts
+++ b/packages/@livestore/common-cf/src/do-rpc/client.test.ts
@@ -1,9 +1,25 @@
 import { Vitest } from '@livestore/utils-dev/node-vitest'
-import { Effect, Fiber, Rpc, RpcClient, RpcGroup, Schema, Stream } from '@livestore/utils/effect'
+import {
+  Effect,
+  Exit,
+  Fiber,
+  Rpc,
+  RpcClient,
+  RpcGroup,
+  type RpcMessage,
+  RpcSerialization,
+  Schema,
+  Stream,
+} from '@livestore/utils/effect'
 
 import type * as CfTypes from '../cf-types.ts'
 import { layerProtocolDurableObject } from './client.ts'
 import { toDurableObjectHandler } from './server.ts'
+
+const EchoRpc = Rpc.make('Echo', {
+  payload: Schema.Struct({ text: Schema.String }),
+  success: Schema.Struct({ echo: Schema.String }),
+})
 
 class Rpcs extends RpcGroup.make(
   Rpc.make('BigStream', {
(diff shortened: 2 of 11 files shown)


IGassmann:
Superseded by https://github.com/livestorejs/livestore/pull/1389

Issue #1437, opened 2026-07-18
Prevent client event loss during orderly Store shutdown


schickling:
## Problem

`Store.shutdown()` can interrupt `ClientSessionSyncProcessor` while a leader push is in flight or while additional events remain queued. Those client-session events are not restart-replayable until the leader accepts them, so orderly shutdown can currently report completion after losing admitted commits.

PR #1405 attempted to fix the loss inside scope finalization. Its review-driven handling of pull teardown, rebase ownership, rejection recovery, and terminal failure added too many interacting states to an already sensitive processor, so that implementation is being reverted for redesign.

## Goal

Restore durable orderly shutdown quickly with an explicit, narrow lifecycle boundary whose success means every admitted event was accepted by the leader—or whose failure is reported without claiming durability.

## Acceptance criteria

- In-flight and queued leader pushes finish serially and preserve FIFO order and configured batch bounds.
- Shutdown closes new admission before draining.
- Pull/rebase ownership cannot clear or replace drain work after shutdown takes ownership.
- Rejected or fatal leader pushes fail the drain; shutdown never reports success after losing pending events.
- Failed/aborted shutdown interrupts blocked work instead of hanging.
- A caller timeout stops waiting without cancelling the underlying cleanup.
- Deterministic tests use explicit synchronization rather than scheduler yields, zero-duration clock nudges, or wall-clock sleeps.
- The implementation keeps shutdown sequencing explicit and substantially smaller than #1405.

## Proposed direction

Add an explicit processor shutdown operation invoked before lifetime-scope teardown:

1. mark the processor closing and reject new admission;
2. stop/join pull so it cannot race queue ownership;
3. close the push queue and let its sole worker drain bounded batches;
4. propagate rejection/fatal failure;
5. close the lifetime scope only after the processor reaches a terminal outcome.

Do not attempt rejection/rebase recovery during shutdown initially. A rejection should fail shutdown; normal recovery remains an open-session behavior.

## Test status

The stricter regression and property tests from #1405 remain in the revert as skipped tests whose source TODO links here. Re-enable them incrementally as the replacement implementation satisfies each invariant.

## References

- Original merged implementation: #1405
- Revert retaining skipped specifications: #1439
- Reopened design/implementation draft: #1440
- Broader lifecycle exploration: #1426
- Related shutdown/hibernation work: #1338

PR #1440, opened 2026-07-18
fix(common): gracefully drain leader pushes on shutdown

schickling:
## Problem

Orderly `Store.shutdown()` can interrupt an in-flight client-to-leader push and abandon additional queued events. PR #1405 fixed that loss, but its handling of scope-finalizer order, rebase ownership, rejection recovery, and terminal failures grew too complex for this core sync path and was auto-merged before explicit human sign-off.

PR #1439 reverts that implementation while retaining its stronger tests as skipped specifications. GitHub cannot literally reopen a merged PR, so this stacked draft recreates #1405 as the active design and review surface.

## Goal

Keep the original problem, implementation, review history, and test discoveries available while we replace the implementation quickly with a smaller explicit pre-scope shutdown boundary tracked by #1437.

This draft is not proposed for merge in its current form.

## Decisions

- Reapply #1405 exactly on top of the revert branch; the resulting tree matches current `main` byte-for-byte.
- Preserve the original implementation here for comparison and review rather than editing history or hiding the rejected approach.
- Use #1437 as the acceptance and redesign authority.
- Prefer an explicit processor shutdown operation before lifetime-scope teardown. Initially fail shutdown-time rejection rather than modeling rebase recovery inside finalizers.

## Verification

- `git diff origin/main HEAD` is empty: this draft exactly recreates the currently merged #1405 tree.
- Original #1405 CI was green across unit, type, lint, integration, performance, examples, and docs.
- PR #1439 proves the primary retained regression is red after revert and keeps 12 shutdown specifications skipped and linked to #1437.

## Complexity

The current implementation adds 108 net production lines to `ClientSessionSyncProcessor` and introduces rejection revision/prefix tracking, drain-failure side channels, finalizer-order coupling, and uninterruptible rebase transitions. Reducing that state and interaction budget is the purpose of reopening the design.

## Concerns

- Do not merge this draft as-is after #1439; it would simply restore the disputed implementation.
- Several tests encode scheduler/virtual-clock coordination that should be replaced with explicit lifecycle handshakes when re-enabled.
- The replacement remains urgent because #1439 intentionally restores known event loss.

## Friction & bottlenecks

- A merged GitHub PR cannot be reopened, requiring this stacked recreation.
- No product-performance bottleneck was introduced or measured.

## Follow-ups

- Implement the narrower lifecycle boundary and re-enable #1437's skipped specifications incrementally.
- Reassess and close or reshape broader lifecycle draft #1426 after the narrow fix is agreed.

## References

- Recreates #1405
- Stacked on revert #1439
- Fix design and tests: #1437
- Broader exploration: #1426


IGassmann (commented):
This PR contains no changes.

PR #1472, opened 2026-07-22
Fix browser rebase rollback metadata retention

slashv:
## Problem

A browser session sends a local event to its leader as `LiveStoreEvent.Client.Encoded`. This is intentional: the session changeset belongs to the session and should not cross the worker RPC boundary. The leader must materialize the event in its own database and create its own changeset.

After accepting a local event, the leader retains it as pending until the sync backend confirms or rebases it. Previously, the leader published that pending state, notified sessions, and queued the backend push before materializing the event. It then materialized a separate `newEvents` instance, so the leader-generated changeset never reached the event retained in pending state.

If a conflicting backend event later caused a rebase, the leader could roll back its own database from its persisted changeset. But the rollback event sent to the browser session came from pending state and had no changeset. The session could not undo its earlier local materialization before applying the rebased event, which could produce a duplicate primary-key error.

Direct in-process profiles masked this because metadata could survive through shared object references. Serialized browser workers exposed the missing leader metadata.

## Solution

The accepted local-push flow is now:

1. Merge and validate the local events.
2. Take the newly appended events from the candidate pending state.
3. Materialize those pending event instances in the leader database and eventlog.
4. Publish the new sync state and notify sessions.
5. Enqueue metadata-free global events for the sync backend.

This puts the leader-generated rollback changeset directly on the events that a later rebase will use. It moves the existing materialization; it does not add another materialization, copy metadata between event instances, or change either RPC boundary.

The new ordering also means sessions and the backend-push fiber cannot observe an accepted leader state unless leader materialization has succeeded. Backend acceptance remains asynchronous, and a conflicting backend event still follows the normal rollback and rebase flow.

## Validation

- Added a LeaderSyncProcessor regression proving that retained and published pending events contain the leader-generated changeset rather than a source-session changeset.
- Focused regression — 1 passed.
- `pnpm exec vitest run tests/package-common/src/leader-thread/LeaderSyncProcessor.test.ts` — 18 passed, 1 skipped.
- `pnpm exec tsc --build packages/@livestore/common/tsconfig.json tests/package-common/tsconfig.json --pretty false` — passed.
- `pnpm exec changeset status --since upstream/main` — passed.
- Targeted `oxfmt --check` and `git diff --check` — passed.

A full automated browser rebase regression would require extending the existing boot-focused browser fixture with commit and backend-conflict controls. This bug was discovered while developing the scenario-testing approach, and the solution was confirmed against the original scenario in a real browser session.

### DIFF
--- a/.changeset/fix-browser-rebase-metadata.md
+++ b/.changeset/fix-browser-rebase-metadata.md
@@ -1,0 +1,5 @@
+---
+'@livestore/common': patch
+---
+
+Fix browser rollback and rebase after conflicting backend events by retaining leader-generated rollback metadata before publishing accepted local events.
--- a/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
+++ b/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
@@ -372,10 +372,19 @@ export const make = Effect.fnUntraced(function* ({
           }
         }
 
+        // A local push appends accepted events after the existing pending events. Materialize that
+        // retained suffix so leader rollback metadata stays on the canonical events used by a later rebase.
+        const newPendingEvents = mergeResult.newSyncState.pending.slice(syncState.pending.length)
+        if (newPendingEvents.length !== mergeResult.newEvents.length) {
+          return yield* Effect.dieDebugger('Local push events must be retained in pending state')
+        }
+
+        yield* materializeEventsBatch({ batchItems: newPendingEvents, deferreds })
+
         yield* SubscriptionRef.set(syncStateSref, mergeResult.newSyncState)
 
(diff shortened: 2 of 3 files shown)


slashv:
I had to revise this PR because an earlier version assumed the event changeset should be sent to the leader from the session but I believe that assumption was wrong. The leader should always save its own changeset based on its current DB state. The key fix remains valid in the change of order of operations in the leader and then storing that changeset.

slashv:
Update once again to what I think is a more elegant solution of materializing the pending events directly so no copy of changeset is necessary and the changeset is attached to the stored pending events instead of the transcient newEvetns.

IGassmann (approved):

Issue #641, opened 2025-09-10
Parent event validation fails during rebase operations in LeaderSyncProcessor


schickling:
# Parent event validation fails during rebase operations in LeaderSyncProcessor

## Problem Description

The integration tests are failing in CI with the error:
```
Parent event X,Y does not exist in eventlog
```

This error occurs during the parent event validation check in `insertIntoEventlog` when `LS_DEV` is enabled.

## Root Cause Analysis

The issue occurs during rebase operations in the `LeaderSyncProcessor` when handling concurrent event streams from multiple clients. Here's the detailed sequence of events that leads to the failure:

### 1. Initial State
- Client A pushes events that get materialized and inserted into the eventlog
- These events have parent-child relationships (e.g., `e2` has parent `e1`)

### 2. Rebase Triggered
- During a pull from the sync backend, the leader receives upstream events
- The `SyncState.merge` operation determines a rebase is needed due to concurrent changes
- This results in:
  - `rollbackEvents`: Events that need to be rolled back
  - `pending`: Rebased events with new sequence numbers that need to be re-applied

### 3. Rollback Execution
- The `rollback()` function is called, which:
  - Reverts the database state by applying inverse changesets
  - **Deletes the rolled-back events from the eventlog** (DELETE FROM eventlog WHERE ...)
  - Deletes the corresponding changeset metadata

### 4. The Problem
- After rollback, the rebased pending events still contain `parentSeqNum` references to the events that were just deleted
- These rebased events are placed in the `syncBackendPushQueue` but are **not re-inserted into the eventlog**
- Later, when new events are pushed that reference these pending events as parents, the validation check fails because it can't find them in the eventlog

## Code Flow

```typescript
// In LeaderSyncProcessor.ts, during rebase:
if (mergeResult._tag === 'rebase') {
  // 1. Filter and queue rebased events for backend push
  const globalRebasedPendingEvents = mergeResult.newSyncState.pending.filter(...)
  yield* restartBackendPushing(globalRebasedPendingEvents)
  
  // 2. Rollback removes events from eventlog
  if (mergeResult.rollbackEvents.length > 0) {
    yield* rollback({
      dbState: db,
      dbEventlog,
      eventNumsToRollback: mergeResult.rollbackEvents.map((_) => _.seqNum),
    })
    // PROBLEM: Rebased pending events are not re-inserted into eventlog here
  }
  
  // 3. Continue with client session updates...
}
```

## Example Scenario

```
Initial eventlog:
[...]

PR #1479, opened 2026-07-25
refactor: use composable SQLite savepoints

IGassmann:
## Problem

Leader and client-session rebases apply multiple inverse SQLite changesets incrementally. If one rollback step fails or the Effect is interrupted, the state database can be left partially rolled back. A regular `BEGIN TRANSACTION` is not composable because these paths may already run inside a transaction.

## Solution

* add synchronous and Effect-based SQLite savepoint helpers
* make Effect cleanup uninterruptible while preserving interruption for the wrapped operation
* validate savepoint names before interpolating them as SQLite identifiers
* wrap leader changeset application and rollback-record deletion in one savepoint
* wrap client-session changeset rollback in one savepoint and mark rollback data as consumed only after release
* document the rebase critical-section invariant in the intent layer
* add focused coverage for commit, failure, nested transaction, and interruption behavior

The eventlog is a separate SQLite database, so its deletion remains a coordinated write after the state savepoint rather than part of the same atomic SQLite operation.

This is an internal refactor and carries an empty changeset.

## Validation

* `vitest run tests/package-common/src/sqlite-db-helper.test.ts tests/package-common/src/client-session/ClientSessionSyncProcessor.test.ts tests/package-common/src/leader-thread/LeaderSyncProcessor.test.ts`
  * 3 files passed
  * 34 tests passed, 13 skipped
* `mono ts`
  * passed
* `oxfmt --check` on all changed files
  * passed
* `mono test unit`
  * relevant common, LiveStore, SQLite, and package-common suites passed
  * Docker/Wrangler suites that require Docker or local listening sockets reported environment failures in the sandbox

The mandated `devenv tasks run lint:full:fix` and `devenv tasks run test:run` entry points could not evaluate because the local effect-utils task module does not define `gh-labels`, which is referenced by `devenv.nix`. The pre-commit `check-quick` hook is blocked by the same unrelated environment error.

### Demo

```text
SAVEPOINT livestore_materialization_rollback
  apply inverse changesets in reverse order
  delete consumed rollback records
RELEASE

failure or interruption
  -> ROLLBACK TO SAVEPOINT
  -> RELEASE SAVEPOINT
  -> re-emit the original cause
```

## Related issues

* Extracted from livestorejs/livestore#1299
* Related rebase rollback context: livestorejs/livestore#641
* Related changeset-apply failure context: livestorejs/livestore#998

### DIFF
--- a/.changeset/empty-composable-sqlite-savepoints.md
+++ b/.changeset/empty-composable-sqlite-savepoints.md
@@ -1,0 +1,4 @@
+---
+---
+
+No release impact. Refactors rebase rollback to use composable SQLite savepoints.
--- a/context/02-system/03-sync/02-processors/spec.md
+++ b/context/02-system/03-sync/02-processors/spec.md
@@ -56,8 +56,10 @@ backend ──pull stream──▶ onNewPullChunk (precedence via semaphore)
   current pending, offers the payload to session pull queues, and persists
   sync metadata for confirmed events; rebase additionally rolls back
   state+eventlog rows and re-seeds pushing from rebased pending
-  (`:466-516`). Backend head advances via `Eventlog.updateBackendHead`
-  (`:462-464`).
+  (`:466-516`). State changeset application and changeset-row deletion are
+  enclosed in a composable SQLite savepoint; eventlog deletion remains a
+  coordinated write to the separate eventlog database. Backend head advances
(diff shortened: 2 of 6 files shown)


IGassmann:
Superseded by github.com/livestorejs/livestore/pull/1502

PR #1488, opened 2026-07-25
WIP: Remove Client Document API

TheAndrewJackson:
> [!IMPORTANT]
> This is a WIP. I'm using the PR interface to leave comments 

## Problem

What problem does this pull request address?

## Solution

How does this change resolve the problem? Mention trade-offs when relevant. Did you consider any other solution ideas?

## Validation

What did you do to verify the change (tests, manual checks, or "not run" with reasoning)?

### Demo (optional)

Share logs, screenshots, CLI output, or quick diagrams (Mermaid/ASCII) that illustrate the change when it helps reviewers grok the impact quickly.

## Related issues

- #...


packages/@livestore/common/src/schema/state/sqlite/table-def.ts
```diff
@@ -38,30 +36,23 @@ export type TableDef<
...
   indexes: SqliteDsl.Index[]
 }>
 
 export namespace TableDef {
-  export type Any = TableDef<any, any>
-}
-
-export type TableOptions = {
-  /** Derived based on whether the table definition has one or more columns (besides the `id` column) */
-  readonly isClientDocumentTable: boolean
```
TheAndrewJackson, line 64: `TableOptions` only contained `isClientDocumentTable` so it's no longer useful not that Client Documents are gone

Issue #1462, opened 2026-07-19
[Bug]: lost DO-RPC push echo can permanently wedge a Cloudflare replica


samducker:
## Summary

A Cloudflare LiveStore replica can become permanently wedged when the sync backend accepts a push but the corresponding echo on the pull stream is lost during a Durable Object restart, eviction, or hibernation transition.

The replica keeps the accepted event pending and replays the same stale-parent batch. The backend is now ahead, so every retry returns `ServerAheadError`. The leader push fiber then parks on `Effect.never`, expecting pull to advance it, but the pull subscription that carried the echo is gone and the stale-parent path does not send a catch-up chunk. All later server-side commits queue behind that push.

We observed this in production on 2026-07-18 with `@livestore/*@0.4.0`. Recovery required rebuilding the replica's persisted state.

This overlaps with #1394's terminal stale-parent state, but the initiating failure here is an accepted push whose pull-channel confirmation is lost across the DO-RPC lifecycle. It also connects #1415, #1111, and the hibernation work in #1338.

## Production evidence

A server-side replica pushed one event with parent 519:

1. The sync backend accepted it as sequence 520.
2. A separate mobile replica later pulled 520 and successfully appended sequences 521 through 533, proving that 520 was persisted.
3. The original replica never observed its echo and retained the event as pending.
4. It replayed the identical 519→520 push at 17:50, 18:40, and 18:55 UTC, spanning two deployments, while the backend remained at head 533.
5. Each replay received `ServerAheadError`; later server-side writes could not pass the parked push.

The accepted event was not lost. The acknowledgement path was.

## Failure chain

The same structural chain is still visible on current `main` at `baaff957`:

1. The sync-cf backend keeps [`rpcSubscriptions` only in memory](https://github.com/livestorejs/livestore/blob/baaff9572d2ff8ddae4d0d89ece130541cbc7f77/packages/%40livestore/sync-cf/src/cf-worker/do/layer.ts#L116). A backend DO restart/eviction silently removes the subscription that would deliver the push echo.

[...]

Issue #1418, opened 2026-07-16
Server DO does not tear down live pull streams on client Interrupt (no Exit emit; DO-RPC path is TODO)


schickling-assistant:
## Problem

On the server DO, cleanup of a live pull stream when the client interrupts is only partially implemented, and the DO-RPC path is entirely a TODO.

WebSocket `Interrupt` removes the request id from the attachment but does not emit an `Exit` stream message:

https://github.com/livestorejs/livestore/blob/8dc44e65982082216a41e1cd1ed995177131ae76/packages/@livestore/sync-cf/src/cf-worker/do/durable-object.ts#L126-L137

```ts
} else if (request._tag === 'Interrupt') {
  // ... remove requestId from pullRequestIds ...
  // TODO also emit `Exit` stream RPC message
}
```

The DO-RPC transport has no interrupt handling at all:

https://github.com/livestorejs/livestore/blob/8dc44e65982082216a41e1cd1ed995177131ae76/packages/@livestore/sync-cf/src/cf-worker/do/pull.ts#L18-L20

```ts
// DO RPC:
// - Further chunks will be emitted manually in `push.ts`
// - If the client sends a `Interrupt` RPC message, TODO
```

## Why it matters

These are acknowledged TODOs in the live-subscription teardown path. The open consequence is that on an abnormal client disconnect the server may retain live-subscriber state / stream resources that are never released. This wants investigation to confirm whether (and how much) resource accumulation actually occurs.

<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_name` | 🥬 cl2-bay |
| `agent_session_id` | 4cc5e6e4-d4c3-4131-9925-d569318b1cf4 |
| `agent_tool` | Claude Code |
| `agent_tool_version` | 2.1.206 |
| `agent_runtime` | Claude Code 2.1.206 |
| `agent_model` | claude-fable-5 |
| `runtime_profile` | /nix/store/qq3avrif77r90ypqbdv3hgd3gvwj5s32-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/pjlb3cwf453ghzmc8jj4v8h29sw6p4dg-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | livestore/schickling/2026-07-15-intent-design |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@4b8e1c5 |
</details>

[...]

Issue #1415, opened 2026-07-16
DO-RPC sync client silently drops live pull update when queue is missing (hibernation branch)


schickling-assistant:
## Problem

In the DO-RPC sync client, when a live pull update arrives for a request whose stream queue is no longer registered (the "DO was hibernated" branch), the update is only logged and never applied:

https://github.com/livestorejs/livestore/blob/8dc44e65982082216a41e1cd1ed995177131ae76/packages/@livestore/sync-cf/src/client/transport/do-rpc-client.ts#L189-L197

```ts
const pullStreamQueue = requestIdQueueMap.get(decodedPayload.requestId)
if (pullStreamQueue === undefined) {
  // Case: DO was hibernated, so we need to manually update the store
  yield* Effect.log(`No pull stream queue found for ${decodedPayload.requestId}`)
} else {
  // Case: DO was still alive, so the existing `pull` will pick up the new events
  yield* Queue.offer(pullStreamQueue, decoded)
}
```

The comment says "we need to manually update the store", but the branch does nothing beyond logging — the decoded `PullResponse` is discarded.

## Why it matters

Under DO hibernation, a live pull update pushed via the RPC callback is silently dropped instead of applied. Whether this is permanent data loss or eventually recovered on a subsequent pull/reconnect depends on the reconnect behavior, which is not traced here — but as written, the affected update never reaches the store through this path.

Related: the queue lookup relies on the module-global `requestIdQueueMap`, see the sibling issue about its scoping.

<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_name` | 🥬 cl2-bay |
| `agent_session_id` | 4cc5e6e4-d4c3-4131-9925-d569318b1cf4 |
| `agent_tool` | Claude Code |
| `agent_tool_version` | 2.1.206 |
| `agent_runtime` | Claude Code 2.1.206 |
| `agent_model` | claude-fable-5 |
| `runtime_profile` | /nix/store/qq3avrif77r90ypqbdv3hgd3gvwj5s32-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/pjlb3cwf453ghzmc8jj4v8h29sw6p4dg-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | livestore/schickling/2026-07-15-intent-design |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@4b8e1c5 |
</details>

PR #1542, opened 2026-08-06
fix(sync-cf): persist DO-RPC subscriber registry across backend reconstruction

bohdanbirdie:
## Problem

The sync backend Durable Object keeps its **DO-RPC** subscriber registry (clients that are themselves Durable Objects, syncing over DO-to-DO RPC) only in memory. When the DO is evicted and rebuilt — deploy, eviction, hibernation — the registry is gone, so the rebuilt instance fans out live events to no DO-RPC subscribers: silent live-sync loss. Separately, the registry was keyed by `storeId`, but one backend serves many client DOs, so a second subscriber to the same store overwrote the first. WebSocket clients are unaffected. #1541 reproduces the reconstruction case as a failing test.

## Solution

- Key the registry by the client's durable-object id (not `storeId`), so multiple DO-RPC subscribers to one store coexist.
- Persist it in the DO's synchronous KV storage (`ctx.storage.kv`, backed by SQLite — schemaless, no migration) as the **single source of truth**: write on subscribe, read on fan-out. No in-memory copy, so the two can't drift.
- The rebuilt backend reads the surviving entries and resumes fan-out.

Each subscriber is plain data (`{ storeId, requestId, callerContext }`) — no live objects — so persisting is just saving who to push to and how.

Out of scope: removing subscribers. Neither the graceful `Interrupt` teardown (#1418) nor reaping ungracefully-abandoned entries (a separate, measurement-gated follow-up) is implemented here — the registry is written and read, never deleted.

## Validation

`cd tests/sync-provider`:

- `TEST_SYNC_PROVIDER=cf-do-rpc-do vitest run src/do-rpc-server-reconstruction.test.ts` → **green** (was red on #1541)
- `TEST_SYNC_PROVIDER=cf-do-rpc-do vitest run src/do-rpc-hibernation.test.ts` → green
- `TEST_SYNC_PROVIDER=cf-ws-do vitest run src/do-hibernation.test.ts` → green
- `tsgo --build`, type-aware oxlint, oxfmt → clean

No public API change (independently audited): the registry lives on the internal `DoCtx` service, `RpcSubscription` is unchanged, and the only public delta is an additive const. No wire/schema change, no user config change, no migration — KV starts empty and refills as clients subscribe.

## Demo

```text
Before                                    After
subscribe ──▶ in-memory Map               subscribe ──▶ ctx.storage.kv.put("rpc-sub:<doId>")
evict + rebuild ──▶ Map = { }             evict + rebuild ──▶ KV survives
push ──▶ fan-out to nobody       ✗        push ──▶ read KV ──▶ fan-out resumes    ✓
```

## Trade-offs and follow-up

- No removal path yet: entries are written on subscribe and never deleted, so abandoned DO-RPC subscribers linger in KV and get a reverse-RPC on every push. Graceful `Interrupt` teardown is tracked in #1418; reaping ungracefully-abandoned entries (hibernated/dead clients that never send `Interrupt`) is a distinct follow-up.
- Client-side next hop, not addressed here: with the callback now firing, a client DO that hibernated still drops the update because its own in-memory routing map is empty — #1415.
- Fan-out now does one local `ctx.storage.kv.list` per push instead of iterating memory — a microsecond-scale SQLite read, dwarfed by the reverse-RPCs already on that path.

## Related

- **#1338** — superseded end-to-end hibernation PR (closed) that first bundled this as "Bug 2 / persist the registry"; this is its reworked server half (it used a SQLite table; this uses `ctx.storage.kv`).
- **#1435** — the DO-RPC hibernation guard test explicitly deferred this exact bug ("the rebuilt DO drops its in-memory subscribers … tracked separately"); this PR is that deferred fix.
- **#1328** — the hibernation fix ("Bug 1", closed) that made this reconstruction loss ("Bug 2") routine; causal origin, not the same bug.
- **#1462** — removes the server-side leg of that production wedge (its failure-chain step 1: the in-memory subscriber `Map` wiped on reconstruction). The downstream stale-parent/`ServerAheadError` deadlock and the client-side leg remain out of scope.
- **#1418** — the graceful-`Interrupt` teardown path for DO-RPC subscribers is still a TODO, so the now-persistent registry is written but never deleted; persisting makes that gap load-bearing.

### DIFF
--- a/packages/@livestore/sync-cf/src/cf-worker/do/push.ts
+++ b/packages/@livestore/sync-cf/src/cf-worker/do/push.ts
@@ -9,6 +9,8 @@ import {
   type Env,
   type ForwardedHeaders,
   type MakeDurableObjectClassOptions,
+  rpcSubscriptionKeyPrefix,
+  type RpcSubscription,
   type StoreId,
   WebSocketAttachmentSchema,
 } from '../shared.ts'
@@ -39,7 +41,7 @@ export const makePush =
   (pushRequest: Omit<SyncMessage.PushRequest, '_tag'>) =>
     Effect.gen(function* () {
       // yield* Effect.log(`Pushing ${decodedMessage.batch.length} events`, decodedMessage.batch)
-      const { backendId, storage, currentHeadRef, updateCurrentHead, rpcSubscriptions } = yield* DoCtx.DoCtx
+      const { backendId, storage, currentHeadRef, updateCurrentHead } = yield* DoCtx.DoCtx
 
       if (pushRequest.batch.length === 0) {
         return SyncMessage.PushAck.make({})
@@ -166,9 +168,11 @@ export const makePush =
           yield* Effect.logDebug(`Broadcasted to ${connectedClients.length} WebSocket clients`)
         }
 
-        // RPC broadcasting would require reconstructing client stubs from clientIds
-        if (rpcSubscriptions.size > 0) {
-          for (const subscription of rpcSubscriptions.values()) {
+        // Subscribers live in the DO's KV storage (single source of truth), so fan-out survives reconstruction.
+        // emitStreamResponse reconstructs each client stub from its callerContext.
+        const rpcSubscriptions = Array.from(ctx.storage.kv.list<RpcSubscription>({ prefix: rpcSubscriptionKeyPrefix }))
+        if (rpcSubscriptions.length > 0) {
+          for (const [, subscription] of rpcSubscriptions) {
             for (const { encoded } of responses) {
               yield* emitStreamResponse({
                 callerContext: subscription.callerContext,
@@ -179,7 +183,7 @@ export const makePush =
             }
           }
 
-          yield* Effect.logDebug(`Broadcasted to ${rpcSubscriptions.size} RPC clients`)
+          yield* Effect.logDebug(`Broadcasted to ${rpcSubscriptions.length} RPC clients`)
         }
       }).pipe(
         Effect.tapCauseLogPretty,
--- a/.changeset/some-suits-tease.md
+++ b/.changeset/some-suits-tease.md
@@ -1,0 +1,5 @@
+---
+"@livestore/sync-cf": patch
+---
+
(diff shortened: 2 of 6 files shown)


packages/@livestore/sync-cf/src/cf-worker/do/push.ts
```diff
@@ -166,9 +168,11 @@ export const makePush =
...
         }
 
-        // RPC broadcasting would require reconstructing client stubs from clientIds
-        if (rpcSubscriptions.size > 0) {
-          for (const subscription of rpcSubscriptions.values()) {
+        // Subscribers live in the DO's KV storage (single source of truth), so fan-out survives reconstruction.
+        // emitStreamResponse reconstructs each client stub from its callerContext.
+        const rpcSubscriptions = Array.from(ctx.storage.kv.list<RpcSubscription>({ prefix: rpcSubscriptionKeyPrefix }))
+        if (rpcSubscriptions.length > 0) {
+          for (const [, subscription] of rpcSubscriptions) {
```
bohdanbirdie, line 175:
```ts
// RPC subscription storage (TODO refactor)
export type RpcSubscription = {
  storeId: StoreId
  payload?: Schema.Json
  subscribedAt: number
  /** Effect RPC request ID */
  requestId: string
  callerContext: {
    bindingName: string
    durableObjectId: string
  }
}
```

@bohdanbirdie look into how this is serialized for the KV storage to make sure it's not corrupted

schickling (approved):

packages/@livestore/sync-cf/src/cf-worker/do/push.ts
bohdanbirdie: Checked — serialization is safe. DO storage uses structured clone (not `JSON.stringify`), and `payload` is decoded through `Schema.Json` before the `put`, which is stricter than structured clone. Nothing gets corrupted.

The read side is an unchecked cast on an unversioned prefix, so changing `RpcSubscription`'s shape later would need a versioned key. Not worth doing now — leaving it as a known surface to revisit if the shape ever changes.

PR #1544, opened 2026-08-07
refactor(sync-cf): scope DO-RPC client pull routing per DO instance

bohdanbirdie:
## What this is

**Preparation** for the DO-RPC client reconstruction fix. It scopes the client's live-pull routing per Durable Object instance, brings the whole public surface (API, examples, docs) along, and adds a regression test that documents the current drop. **This is not the behavioural fix yet** — a reconstructed client still drops the in-flight update; the follow-up recovery change makes it survive.

## Problem

A **DO-RPC** subscriber (a LiveStore client that is itself a Durable Object, syncing over DO-to-DO RPC) loses a live update when the client DO is evicted and rebuilt. Its live-pull routing map was a module global. Cloudflare discards in-memory state — [globals included](https://developers.cloudflare.com/durable-objects/reference/in-memory-state/) — on hibernation/eviction, and may leak it across co-located DOs, so a reconstructed instance either reuses a stale entry or leaks one, and the backend's *reverse-RPC* push (the backend calling `syncUpdateRpc` on the client DO) is misrouted or dropped. #1541 / #1542 fixed the mirror-image loss on the server.

## Public API change (breaking for DO-RPC clients)

- `handleSyncUpdateRpc(payload)` → `handleSyncUpdateRpc(ctx, payload)` — pass the client DO's `ctx`.
- `makeDoRpcSync` gains a `durableObjectState` option.

Routing now lives in a `WeakMap` keyed by the DO's `ctx`, so it resets on reconstruction and can't bleed across co-located instances. Also fixes a routing-map queue leak (entries are deleted on pull-stream release).

**Surface updated:** examples (`cloudflare-todomvc`, `web-email-client` Mailbox + Thread), docs code snippets, the integration worker, and the public reference pages (`sync-providers/cloudflare`, `platform-adapters/cloudflare-durable-object-adapter`).

## The test

A store-based reconstruction test under the `cf-do-rpc-do` provider (`wrangler dev`, real-idle eviction): a new `StoreClientDo` boots a real `livePull` subscriber store, applies + persists an event pushed *before* reconstruction, is idled into a genuine eviction (instance id changes) while the server stays warm, then a heal-free eventlog probe shows the *after* event was dropped while a fresh catch-up subscriber still receives it from the server. It asserts the current DROP (`not.toContain('after-reconstruct')`); the follow-up recovery change flips it to `toContain`.

**Notes:**
- **Why a new `StoreClientDo` (not reuse `TestClientDo`)** — `TestClientDo` is the WS→DO relay every `cf-do-rpc` test rides on; it forwards RPCs but never boots a store, so it has no eventlog to read. Converting it would break every relay-dependent test.
- **Heal-free read** — the probe reads the persisted `eventlog` raw and boots no store; any store-level read would trigger a boot catch-up pull that masks the drop (`livePull: false` / `initialSyncOptions: Skip` don't prevent it — only *no backend* does).
- **Alternative tried** — the deterministic, gate-able path is `@cloudflare/vitest-pool-workers` + `evictDurableObject`, but it can't run in this repo yet (a `vite`/`vitest` duplicate stops the workerd runner from collecting). Fell back to the `wrangler dev` real-idle harness.

## Where the routing map lives (options considered)

*Where the map is stored / how it's linked to the instance:*

| Option | Where stored / how linked | Verdict |
|---|---|---|
| module global (baseline) | `Map` at module scope; the free function reaches it only as a global | the bug — shared/stale across co-located instances, survives reconstruction locally, entries leak |
| **B — `WeakMap<ctx>`** ✅ | `WeakMap<DurableObjectState, Map<reqId, Queue>>` keyed by `ctx`; `handleSyncUpdateRpc(ctx, payload)` gains a `ctx` param | **chosen** — per-instance, dead entries auto-GC, no cast, no leaked types, no `createStoreDo` shape change, one-line consumer edit |
| G — opaque handle | library handle held as a DO instance property, passed back into `createStoreDo` | runner-up — no leak, but adds an input option + consumer ceremony |
| A — instance field | raw `Map` on the DO (`this.map`) | leaks Effect internals (`Queue<PullResponse>`) into the consumer DO's type |
| C — `Symbol` on `ctx` | `(ctx as any)[sym]` | needs `as any` + mutates a CF-owned object (repo forbids casts) |
| E — return `{ store, syncUpdateRpc }` | closed over inside `createStoreDo` | breaks the "returns a store" contract |
| F — method on `Store` | `store.syncUpdateRpc(...)` | wrong layer — `Store` is cross-environment (browser/node/expo) |

## Follow-up

- **The actual fix (separate PR):** client recovery — a reconstructed, store-less client re-boots its store and catches up on the reverse-RPC wake, turning the drop into an applied update (flips the test assertion to `toContain`).
- Replace the raw-`eventlog` probe with a backend-less-store + store-API read to drop the persistence-format coupling (needs `createStoreDo` to not require a `syncBackendStub`).

## Related

- #1415 — the client-side dropped-update report this scopes/characterizes.
- #1541 / #1542 — the mirror-image server-side loss (test + fix); this is the client counterpart.

### DIFF
--- a/.changeset/scope-do-rpc-client-pull-routing.md
+++ b/.changeset/scope-do-rpc-client-pull-routing.md
@@ -1,0 +1,5 @@
+---
+"@livestore/sync-cf": minor
+---
+
+**Cloudflare DO-RPC client:** Scope the client's live-pull routing to the Durable Object instance. `handleSyncUpdateRpc` now takes the DO's `ctx` (`handleSyncUpdateRpc(ctx, payload)`) and `makeDoRpcSync` takes a `durableObjectState` option, so routing resets when the client Durable Object is reconstructed (deploy, eviction, hibernation) and cannot leak across co-located instances. This also fixes a routing-map queue leak. A reconstructed client still drops the in-flight update until the follow-up recovery change lands (#1415).
--- a/docs/src/content/_assets/code/reference/platform-adapters/cloudflare/client-do.ts
+++ b/docs/src/content/_assets/code/reference/platform-adapters/cloudflare/client-do.ts
@@ -79,6 +79,7 @@ export class LiveStoreClientDO extends DurableObject<Env> implements ClientDoWit
   }
 
   async syncUpdateRpc(payload: Uint8Array<ArrayBuffer>) {
-    await handleSyncUpdateRpc(payload)
+    // @ts-expect-error TODO remove once CF types are fixed in https://github.com/cloudflare/workerd/issues/4811
+    await handleSyncUpdateRpc(this.ctx, payload)
   }
 }
--- a/docs/src/content/_assets/code/reference/syncing/cloudflare/client-do-rpc.ts
+++ b/docs/src/content/_assets/code/reference/syncing/cloudflare/client-do-rpc.ts
@@ -6,6 +6,7 @@ declare const syncBackendDurableObject: CfTypes.DurableObjectStub<SyncBackendRpc
 
 export const syncBackend = makeDoRpcSync({
   syncBackendStub: syncBackendDurableObject,
+  durableObjectState: state,
   durableObjectContext: {
     bindingName: 'CLIENT_DO',
     durableObjectId: state.id.toString(),
--- a/docs/src/content/docs/platform-adapters/cloudflare-durable-object-adapter.mdx
+++ b/docs/src/content/docs/platform-adapters/cloudflare-durable-object-adapter.mdx
@@ -113,7 +113,7 @@ Creates a LiveStore instance inside a Durable Object.
 
 ### `syncUpdateRpc(payload)`
 
-Client Durable Objects must implement this method so the sync backend can deliver live updates. `createStoreDoPromise` wires it up automatically—just forward the payload to `handleSyncUpdateRpc` (see the client Durable Object example above).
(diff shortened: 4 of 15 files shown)


bohdanbirdie:
**Follow-up design review: can the per-instance routing map be avoided?**

Revisited whether the `WeakMap<ctx>` partition (option B) could be dropped for something simpler — specifically making Effect RPC request ids **globally unique** (Effect v4 exposes `RpcClient.make(group, { generateRequestId })`, and `RequestId` is `string | number`, so a UUID is a drop-in) so a single flat module-level `Map<requestId, Queue>` couldn't misroute across co-located DOs.

Conclusion: it doesn't simplify anything — keeping option B.

- In the current design, unique ids **change nothing**. `requestId` is only ever resolved within a single client DO's scope: on the client via `WeakMap<ctx>`, and on the server subscriptions are keyed by `callerContext.durableObjectId`, not by `requestId`. The id never acts as a cross-instance key, so global uniqueness is redundant.
- Unique ids would matter *only* if we removed the outer map and used a flat module map — and even then they'd only restore routing correctness. We'd lose the `WeakMap`'s automatic GC of a dead instance's entries (a flat module map holds strong refs, so it leaks dead queues when a co-located DO is evicted while its isolate stays warm), and we'd take a dependency on an `unstable/` Effect API for id generation.

So `WeakMap<ctx>` gives per-instance scoping **and** free cleanup in one construct, without touching Effect's id generation.

schickling (approved):

Issue #1265, opened 2026-06-01
Epic: repo split — core engine + livestore-contrib


schickling-assistant:
## Epic: LiveStore Core/Contrib Split

Canonical tracking epic for the end-to-end repository split, aggregate witness PR, and release-readiness follow-up work.

## Current heads

- [x] Core split completed on `main` at `ff95d770b9eefa2d66a6a512779a66278f2a9a4a`.
- [x] Contrib bootstrap completed on `main` at `a4e620a966283c4e2f483422937dea499d298409`.
- [x] Core release-flow PR merged: livestorejs/livestore#1325.
- [x] Contrib release-flow PR merged: livestorejs/livestore-contrib#4.
- [x] Contrib snapshot smoke-install fix merged: livestorejs/livestore-contrib#5.
- [x] Contrib npm visibility hardening merged: livestorejs/livestore-contrib#6 at `e941d6af9d68875bd85a6f4573cf73e63a95056f`.

## Completed

- [x] Core packages remain in `livestorejs/livestore`; moved framework integrations, adapters, sync providers, devtools-contrib packages, GraphQL, CLI tooling, and moved examples live in `livestorejs/livestore-contrib`.
- [x] `todomvc-experimental` and `web-todomvc-custom-elements` moved to contrib.
- [x] Contrib owns its moved-package topology; core does not couple `genie/repo-topology.ts` back to contrib.
- [x] Aggregate composition owns cross-repo witness state.
- [x] Contrib CI and repo settings are generated through Genie.
- [x] Live repository settings for both repos have `allow_auto_merge=true` and `delete_branch_on_merge=true`.
- [x] Required contrib checks are managed through generated repo settings.
- [x] Contrib CI uses Namespace runners.
- [x] Genie-generated files are covered by git attributes.
- [x] README and steady-state docs were cleaned up; temporary handoff/planning overhead was removed.
- [x] Core release flow uses `release.yml` as the single npm trusted-publisher workflow for core stable and snapshot package publishes, with manual publish dispatch guarded to `main`.
- [x] Contrib release flow publishes real snapshots from `release.yml`, derives the core snapshot from the pinned core commit in `megarepo.lock`, uses a distinct contrib snapshot version, restores generated files after dry-runs, and smoke-installs packed tarballs in the verified real-publish path.
[...]

PR #1563, opened 2026-08-19
Build the PR snapshot pipeline from a shared factory

schickling:
## Problem

This repo owns a six-job pipeline that publishes an immutable npm snapshot of a pull request's package cohort without granting the PR author any credential — including, since #1560, fork pull requests trusted through the `ci:publish-snapshot` label.

livestore-contrib needs exactly the same guarantee for its own packages. The obvious route is for contrib to copy the pipeline, which would fork the code that decides what may be published: the validator that gates publication, the authorization checks, and the version scheme would drift apart precisely where drift is most expensive.

## Goal

The pipeline moves to a shared factory in effect-utils, and this repo becomes its first consumer. Contrib then composes the same implementation rather than a lookalike.

## Decisions

**Extract here first, and require byte-identical output.** The point of doing the extraction in the repo that already runs this pipeline in production is that equivalence is checkable. Contrib inherits code proven against a real publication rather than a plausible-looking copy.

**The pin bump is a separate commit from the refactor.** The effect-utils pin was roughly a hundred commits behind, so regenerating picks up unrelated drift across seven workflows. Isolating it keeps the refactor commit's diff reviewable — its generated-workflow diff is empty.

**Dropping the local `ci:publish-snapshot` definition rides with the pin bump, not the refactor.** effect-utils now supplies the label via `commonLabels`, and the label normalizer sorts without deduplicating. A tree carrying both definitions emits the entry twice — confirmed by observing the duplicate appear. Splitting them would leave an intermediate commit with a duplicated label.

## Verification

**The refactor changes no generated output.** With the pin bump committed separately, regenerating on top of the refactor gives:

```
git diff --exit-code -- .github/workflows/ci.yml .github/workflows/release.yml .github/labels.json
GATE_EXIT=0
```

690 lines of job definitions move without changing a line of output. Re-confirmed after formatting.

**Presence checked separately**, because an empty diff can equally mean content vanished from both sides: all five release jobs, the pack job, all three dispatch inputs, the promote mode option and the schedule are present in the regenerated output, and the emitted `pr-snapshot-artifact.mjs` is byte-identical to the shared source apart from its two-line generated header.

**The gate was shown to be capable of failing.** Perturbing the shared factory's `packJobId` default produces 6 changed lines; restoring it produces 0. That control also demonstrates why `packJobId` is a single option: one change moves the job key in `ci.yml` and both `jq` name matches in `release.yml` together.

`devenv tasks run check:all` — see Concerns.

## Complexity

No new complexity; net removal here. Six job definitions become two call sites.

## Concerns

`otel:verify:setup` fails locally. It is under investigation and appears to be adoption fallout from the ~100-commit effect-utils bump rather than anything about PR snapshots — the capture reports fewer spans than the verifier expects while the child task exits 0. This PR should not be flipped ready until that is resolved or shown to be pre-existing.

The megarepo and devenv inputs currently point at an effect-utils **branch**, not `main`. They must be retargeted to `main@<merged-sha>` once the effect-utils PR lands, before this merges.

## Friction & bottlenecks

- `oxfmt .` reformats the entire repo: 137 files, 134 of them pre-existing changeset files unrelated to this change. Those were reverted; only files this change actually touches are included.
- Consuming an effect-utils branch requires retargeting three places — `megarepo.kdl`, both `devenv.yaml` inputs, and then the two lockfiles via `mr:apply` — each of which must be undone before merge.
- The generator fingerprints this repo's own sources, and composed members are symlinks that `git ls-files` will not traverse, so editing a shared factory in place locally is silently skipped as up to date. Real consumption is unaffected because a shared change arrives as a tracked lock bump.

## Follow-ups

- Retarget megarepo/devenv inputs to `main` once the effect-utils PR merges.
- Resolve or discharge the `otel:verify:setup` failure.

## References

Depends on overengineeringstudio/effect-utils#1087. Paired with the livestore-contrib PR that consumes the same factory. Refs #1265, #1560, #1562.


schickling:
Diagnosed this branch's CI failure while working the sibling contrib PR (livestorejs/livestore-contrib#40), which composes the same effect-utils pin. Posting the root cause since it is a one-file fix.

**Every job fails at "Resolve devenv" with:**

```
bash: line 7: /home/runner/work/_temp/genie-ci-scripts/resolve-devenv.sh: No such file or directory
```

**Cause.** The effect-utils bump in this PR (`99c8c77…` → `8c686de0…`) changed the workflow templates so the *Validate Nix store* step **sources** `resolve-devenv.sh` instead of inlining `resolve_devenv()`:

- `genie/ci-workflow/setup.ts:997` — `. "${preparedCiRuntimeScriptsDir}/resolve-devenv.sh"` (inside `validateNixStoreStepFor`)
- vs `setup.ts:347` — still inlines it (inside `preparePinnedDevenvStepFor`)

effect-utils ships the script as `ciWorkflowSupportFiles.resolveDevenv` (`genie/ci-workflow/support-files.ts:217`, path `genie/ci-scripts/resolve-devenv.sh`), but **each consumer repo must add its own `.genie.ts` to emit it** — the "Prepare CI helper scripts" step only copies files that exist in `genie/ci-scripts/`. This PR regenerated the workflows onto the sourcing form without adding that generator, so the step copies a directory that has no `resolve-devenv.sh` in it.

Deterministic, not flaky: 7/7 runs on this branch fail identically, and the sibling `ci` run fails the same way in `lint`, `type-check`, `test-unit` and `pack-pr-snapshot`.

**Fix** — add `genie/ci-scripts/resolve-devenv.sh.genie.ts`, mirroring the existing retry-helper generator exactly:

```ts
import { ciWorkflowSupportFiles } from '../../repos/effect-utils/genie/ci-workflow.ts'

export default ciWorkflowSupportFiles.resolveDevenv.output
```

then regenerate. (Adjust the import path to however this repo reaches effect-utils.)

**Why contrib is unaffected**, in case it looks like it should be: contrib composes the no-arg `validateNixStoreStep`, whose emitted YAML does not use the sourcing form — verified `grep -c 'genie-ci-scripts/resolve-devenv.sh'` → 0 across both its workflows, with CI green on the same effect-utils pin. So this is specific to the step variant used here, not the bump itself.

PR #1565, opened 2026-08-20
Verify the minimal development image in PR CI

schickling:
## Problem

The Docker cold-start oracle can drift without detection if it remains only a manual local check.

## Goal

Run the canonical root Dockerfile as an independent pull-request CI job, without first entering Nix/devenv.

## Decisions

- Add the stable `minimal-dev` job without renaming existing gates.
- Use stock `ubuntu-latest`, checkout, and a direct Docker build.
- Keep this layer PR-only and defer required-gate policy to the final stack layer.
- Add no cache, publish step, matrix, or shared environment setup.
- Edit Genie authority and commit the generated workflow together.

## Verification

`devenv tasks run genie:run`, `lint:check:genie`, `genie:check`, `git diff --check`, and commit `check-quick` passed. The unchanged Docker oracle had already passed its uncached build in the parent layer.

## Complexity

One independent two-step generated CI job.

## Concerns

Every PR initially pays for a cold workspace install and image export; caching remains evidence-driven follow-up work.

## Friction & bottlenecks

Repository Genie checks are the focused authority; standalone actionlint also reports unrelated existing custom-runner-label and shellcheck findings.

## Follow-ups

Observe CI timing before adding cache or image-distribution complexity.

## References

- Depends on #1564.
- Layer 2 of stack #1566; followed by #1567, #1568, and #1569.

PR #1567, opened 2026-08-21
Expand the portable development loop

schickling:
## Problem

The cold-start image proves only a narrow batch check and does not yet provide a conventional interactive loop or establish the useful portable capability boundary.

## Goal

Make portable development first-class with a thin Compose entry point and broaden the finite Docker oracle to the representative work that succeeds without adding tools.

## Decisions

- Bind the exclusive checkout into one `development` service; preserve the host UID/GID and use plain Bash as the entry point.
- Add no launcher, devcontainer, named dependency volume, or fixed application-port catalog.
- Derive `WORKSPACE_ROOT` from the shared package location when absent.
- Resolve docs branch links from CI refs, then fall back to `main`, without requiring Git.
- Expand the Docker oracle to TypeScript, the stable common unit suite, a Vite example build, local Wrangler build, and docs check.
- Exclude dev servers, full docs build, Playwright, generators, releases, and Nix from the finite image build.

## Verification

- Uncached expanded Docker build: PASS in 115.445 seconds; 271 common tests plus TypeScript, Vite, Wrangler, and docs check passed.
- `docker compose config`: PASS.
- Disposable checkout Compose shell as UID 1000/GID 100: frozen install, TypeScript, 6 focused tests, and docs check passed.
- `devenv tasks run lint:full:fix`, `git diff --check`, and commit `check-quick`: PASS.

## Complexity

One small Compose service and two local defaulting fixes. Existing commands remain the implementation.

## Concerns

- A bind-mounted checkout has exclusive ownership; simultaneous host/container installs are unsupported.
- Full docs needs a browser, Playwright needs installed browsers, and generator/release/wa-sqlite surfaces remain full-lane work.

## Friction & bottlenecks

The full docs experiment rendered all 158 snippets but failed after 399 seconds because Puppeteer Chrome was absent. That result shaped the portable boundary rather than adding browser installation weight.

## Follow-ups

Image size and caching should be optimized only if observed CI or onboarding costs justify them.

## References

- Depends on #1565.
- Layer 3 of stack #1566; followed by #1568 and #1569.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,7 @@ RUN node --version \
   && pnpm --version \
   && pnpm install --frozen-lockfile \
   && pnpm exec tsc -b packages/@livestore/livestore --pretty false \
-  && pnpm --filter @livestore/common exec vitest run src/schema/EventSequenceNumber.test.ts
+  && pnpm --filter @livestore/common exec vitest run \
+  && pnpm --filter livestore-example-web-todomvc run build \
+  && pnpm --filter livestore-example-cloudflare-todomvc run build \
+  && pnpm --filter @local/docs run check
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,12 @@
+services:
+  development:
+    build: .
+    working_dir: /workspace
+    user: "${LOCAL_UID:-1000}:${LOCAL_GID:-100}"
+    environment:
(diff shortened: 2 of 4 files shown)


schickling:
Superseded by #1564, which now contains the complete executable Minimal Setup layer after the stack was consolidated.

PR #1573, opened 2026-08-21
Expose backend-confirmed delivery in SyncStatus

### LINKED ISSUE #1553: `SyncStatus.isSynced` can be mistaken for backend-confirmed sync
## Problem

`store.syncStatus().isSynced` becomes `true` when the client session has delivered its events to the leader. The leader may still have those events pending confirmation from the sync backend.

The name and documentation examples can therefore be mistaken for end-to-end synchronization.

## Expected

Make the scope unambiguous—either clarify/rename the existing session-to-leader status or expose a distinct backend-confirmed status.

This is not about event loss or access to confirmed events; `store.events()` and `eventsStream()` already expose backend-confirmed events.

slashv:
## Problem

`SyncStatus.isSynced` only reports session-to-leader delivery. A session can therefore report synced while the leader still has events awaiting sync-backend confirmation.

## Solution

Add three flat fields to the existing synchronous status:

- `backendHead`: latest backend-confirmed head observed through the leader
- `backendPendingCount`: leader events still awaiting backend confirmation
- `isBackendSynced`: latest-observed convergence across both propagation boundaries

The Store seeds and maintains a local cache from the leader's existing sync-state getter and stream, so this adds no transport protocol. A session watermark prevents stale cached leader state from producing a false positive. `syncStatusStream()` reacts to changes at either boundary.

The first commit contains deterministic regression tests; the second contains the implementation, derived docs, changelog, changeset, and delta resolution.

## Trade-offs

This is a latest-observed synchronous delivery snapshot. It may briefly remain false while a newer leader observation is in transit, but it must not report true from stale leader state. It is not sync-worker health: converged observed heads may leave `isBackendSynced` true while a terminal worker is parked, and the field does not guarantee future progress.

This PR does not add worker-health state, per-event receipts, or backend delivery-failure classification.

## Validation

- `pnpm exec vitest run packages/@livestore/livestore/src/store/store-sync-status.test.ts` (2 passed)
- `pnpm exec vitest run tests/package-common/src/intent-layer/intent-layer.test.ts` (8 passed)
- `devenv tasks run ts:check`
- `devenv tasks run test:integration:sync-provider:mock`
- `DEVENV_TASK_PASSTHROUGH=1 devenv tasks run docs:build`

## Stack

- Layer 1: #1554 — synchronous Store contract
- Layer 2: this PR — regression tests and implementation

Depends on #1554.
Closes #1553.

Related clarification: #1580.

### DIFF
--- a/.changeset/backend-confirmed-sync-status.md
+++ b/.changeset/backend-confirmed-sync-status.md
@@ -1,0 +1,5 @@
+---
+"@livestore/livestore": minor
+---
+
+Add backend confirmation fields to `SyncStatus` while preserving the existing session-to-leader fields.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
   accepted event cannot advance the backend head without notifying subscribers
   ([#1537](https://github.com/livestorejs/livestore/pull/1537)).
 - Removed redundant devenv package entries now owned by the task guard modules.
+- **Sync status API:** Added `backendHead`, `backendPendingCount`, and
+  `isBackendSynced` to `SyncStatus`, allowing sessions to distinguish leader
+  acceptance from sync-backend confirmation while preserving the existing
+  session-to-leader fields ([#1553](https://github.com/livestorejs/livestore/issues/1553)).
 - **Sync correctness:** Prevented later client-session events from crossing an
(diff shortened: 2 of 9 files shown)


slashv:
@codex

Issue #1532, opened 2026-07-31
`backendHead` still advances non-atomically on pull — process kill (not materializer crash) reproduces #409 on 0.4.0


locandorg:
Follow-up to #409, which was closed as completed on 2025-09-15 after the
materializer-crash repro stopped reproducing. The closing comment was:

> I believe there could still be some scenarios where this could happen but I'd
> need a repro to address them.

This is one of those scenarios, with a repro. The trigger is a **process death**
rather than a materializer failure, so the fix that resolved the original repro
doesn't cover it.

## Still non-atomic in 0.4.0

`@livestore/common@0.4.0`, `dist/leader-thread/LeaderSyncProcessor.js:418`:

```js
const newBackendHead = newEvents.at(-1).seqNum
Eventlog.updateBackendHead(dbEventlog, newBackendHead)   // ← committed immediately
// ... rebase handling, rollback, queue offers ...
// ... events are inserted into the eventlog further down
```

The head write and the event insert are two separate commits with no transaction
spanning them. #409 framed this as "crash in materializer"; the underlying
problem is that **any** interruption between the two commits leaves
`backendHead > localHead` on disk permanently.

## Repro (no faulty code needed)

Unlike #409 this doesn't need a second writer or a deliberately broken
materializer — just a kill at the wrong moment.

1. Open a store with a sync backend and let it settle.
2. From another client, push enough events that the pull is not instantaneous.
3. While that pull is in flight, kill the process (`SIGKILL`, iOS
   suspend-kill, force-quit, or OOM).
4. Relaunch.

Boot dies with:

```
During boot the backend head (N) should never be greater than the local head (M)
```

Equivalent one-liner if you'd rather not race it — this is the exact on-disk
state step 3 produces:

```sql
-- in livestore-eventlog@6.db
UPDATE __livestore_sync_status
SET head = (SELECT MAX(seqNumGlobal) FROM eventlog) + 1;
```

Every subsequent boot fails the same way. It is not self-healing: one kill
during a pull arms the crash for every launch afterwards.

## Impact

We hit this in production on iOS, where the OS suspend-kills backgrounded apps
routinely — so step 3 is not an edge case, it's normal platform behaviour. The
symptom users see is a hard crash ~3.5s after every cold launch, with no
interaction required, and it never clears on its own.

## Suggested fix

Wrap the head update and the event insert in a single transaction so the pair is
atomic. If `trimChangesetRows` (noted in #409) genuinely needs the updated head
[...]

Issue #1394, opened 2026-07-15
[Bug]: sync-cf deadlocks stale-parent pushes after concurrent offline reconnect


schickling-assistant:
## Summary

When two independent clients append while fully offline and reconnect concurrently, one push advances the sync-cf Durable Object head. The other client then pushes with a stale parent and receives `ServerAheadError`. That client never receives the catch-up pull chunk that `LeaderSyncProcessor` requires to resume, so its backend-pushing fiber waits forever and its local events remain stranded.

This is reproducible with current released packages and is still present on LiveStore `main` at eabb8feeff4179cd9cc626f276dd0785ef1c1b61.

Minimal reproduction: https://github.com/schickling-repros/2026-07-livestore-sync-cf-reconnect-deadlock

## Reproduction

```sh
pnpm install
pnpm exec playwright install chromium
pnpm test
```

The repro contains only one synced `Append` event and its SQLite materializer. It runs a local sync-cf Worker and SQLite Durable Object, then:

1. Opens two separate browser contexts with independent OPFS/SharedWorkers.
2. Waits for both clients to connect.
3. Takes both contexts offline and waits for `isConnected === false`.
4. Appends one event in each context.
5. Reconnects both contexts concurrently.
6. Opens a fresh third context to read the backend state.

One observed run after 20 seconds was:

```text
offline: A=["A"], B=["B"]
after reconnect: A=["A"], B=["B"], fresh=["A"]
```

The winning branch can vary with scheduling; the invariant is that the stale-parent writer's event is not pushed and the clients do not converge.

## Expected

The stale writer pulls the backend events after its parent, rebases/retries its queued event, and all three clients converge to `A` and `B`.

## Actual

The stale writer reports `ServerAheadError`, then its push processor waits indefinitely. No catch-up pull chunk is emitted on the rejection path, so nothing interrupts/restarts that push. The event remains only in that client's local store, while a fresh client sees only the branch that reached the backend first.

## Root cause

In the published 0.4.0 source:

- `packages/@livestore/sync-cf/src/cf-worker/do/push.ts:67-82` returns `ServerAheadError` immediately when `firstEventParent !== currentHead`.
- `packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts:905-910` handles `ServerAheadError` with `Effect.never`; its comment explicitly says the backend must emit a new pull chunk alongside the error.
[...]

Issue #1365, opened 2026-07-04
sync-cf ws client deadlocks permanently after server connection loss: resume Ack is silently ignored, push fiber never retries


samducker:
## Summary

If a `@livestore/sync-cf` WebSocket connection drops while streams are active (e.g. a `wrangler deploy` restarts the sync backend DO), the client enters a **permanent sync deadlock** that survives client restarts. We hit this in production: every client stopped syncing at the exact minute of a deploy and stayed dead for 6 days until we patched the server.

Two coupled defects:

### 1. Orphaned resume-Ack is silently ignored (pull deadlock)

After reconnecting, the client's **first frame** on the brand-new connection is an Ack for its old pull stream:

```json
{"_tag":"Ack","requestId":"0"}
```

The new connection/DO has no such request. `@effect/rpc`'s RpcServer ignores Acks for unknown requestIds (`RpcServer.js`, `case \"Ack\": return latch ? latch.open : Effect.void`), so nothing is sent back. The client waits ~5s, times out (`ConnectionError: Sync backend did not respond within 5000ms`), reconnects, and sends the same orphaned Ack again — forever. The resume state persists in the client store, so restarting the app does not recover.

At prod log level the client is silent; server-side there is nothing to log — connections just show as `canceled` after ~1s.

### 2. Push fiber never retries after connection death

Independently, the client's push machinery stays wedged: after we fixed (1) server-side, the same client re-established live pulls (fresh `SyncWsRpc.Pull`, chunks flowing, legitimate Acks) but **never re-sends a push request** for its queued local events — apparently still awaiting a response to the push request issued on the dead connection. Related: #1136 handled push-fiber *failure*; connection death appears not to fail the fiber.

## Environment

- `@livestore/*` 0.4.0, `@effect/rpc` 0.75.1, `effect` 3.21.2
- Sync backend: `makeDurableObject` from `@livestore/sync-cf/cf-worker` on Cloudflare DO (SQLite), ws transport
- Clients: Expo/React Native (adapter-expo) and Node (`@livestore/cli` / adapter-node); both affected

## Reproduction

1. Client connected over ws with an active live pull (and pending local events).
2. Restart the sync backend DO (`wrangler deploy` is enough) so the socket drops.
[...]

PR #1576, opened 2026-08-21
Recover ServerAhead stalls by restarting backend pull

slashv:
## Problem

A backend can accept and persist leader event A1 while its pull publication is lost. After another client advances the backend with B2, the leader can locally admit A2 from stale parent A1 and receive `ServerAheadError`. Parking the push worker on the existing pull can wedge the unresolved prefix forever.

This is the remaining core recovery leg tracked by #1462. It is stacked on #1575, which owns processor decision 0003 and the provider-neutral contract.

## Solution

- Convert `ServerAheadError` into a capacity-one catch-up request and stop the sole push worker, preserving the unresolved-prefix fence.
- Keep one persistent pull owner. It remains available after finite pull completion and re-reads the persisted cursor for every replacement.
- Coordinate retirement with a poison-agnostic canonical-application fence and retirement flag. An application already inside the fence completes; later applications self-stop. Pull cleanup runs outside the fence, then the owner inspects the generation exit.
- Give terminal canonical application failure precedence over replacement, so a later poisoned-event implementation can reuse the boundary without having its failure interrupted or masked.
- Coalesce restart requests and start the replacement only after the old generation has terminated, preserving a maximum of one active pull generation.
- Retry backend pushes only for `IsOfflineError`. `UnknownError` is terminal; `ServerAheadError` is reconciliation rather than transport retry.
- Keep ordinary pull merge as the only path that confirms/rebases the fenced prefix, rebuilds the push FIFO, and resumes propagation.

The mock backend exposes deterministic completion, attempt, cursor, stored-event, and active-pull observations while retaining a subscriber per evaluated live pull.

Trade-off: recovery may replay history from the persisted cursor. The merge core already owns idempotent confirmation/rebase; no provider wire contract or Cloudflare adapter behavior changes.

## Deterministic coverage

- accepted push/lost publication plus stale-parent advance produces a natural `ServerAheadError` and converges;
- a normally completed finite pull is restarted by a later catch-up request;
- a catch-up request arriving after cursor advancement cannot truncate the in-flight canonical application;
- a terminal canonical application failure wins the restart race and no replacement starts;
- `UnknownError` receives exactly one push attempt and surfaces through shutdown;
- replacement pulls replay persisted history through a fresh subscription and never overlap.

## Validation

- `devenv tasks run ts:check`
- `pnpm exec vitest run tests/package-common/src/intent-layer/intent-layer.test.ts` — 8 passed
- `pnpm exec vitest run packages/@livestore/common/src/sync/mock-sync-backend.test.ts` — 4 passed
- `pnpm exec vitest run tests/package-common/src/leader-thread/LeaderSyncProcessor.test.ts` — 23 passed, 1 existing skip
- `devenv tasks run test:unit` — passed
- `devenv tasks run test:integration:sync-provider:mock` — 19 passed, 9 provider-inapplicable skips

The GitHub matrix is rerunning on commit `572870b2e`.

### Recovery sequence

```text
push A1 accepted --X publication
backend advances B2
push A2 -> ServerAhead -> prefix remains fenced
                         -> request catch-up
in-flight pull apply finishes or fails (failure wins)
old pull generation terminates
replacement pull(persisted cursor) replays A1+B2
ordinary merge rebases A2 -> resumed push
```

## Related issues and foundations

- #1462: accepted-push/lost-confirmation incident and core recovery.
- #1394: stale-parent reconnect deadlock.
- #1365: related WebSocket connection-loss incident; orphaned resume-Ack remains out of scope.
- #1530: prefix fencing and explicit leader push reservations.
- #1537: complementary Cloudflare push/publication atomicity.
- #1545: DO-RPC reconstruction recovery through boot catch-up.
- #1532: separate pull crash/cross-database atomicity work.

### DIFF
--- a/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
+++ b/packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
@@ -10,11 +10,13 @@ import {
   Duration,
   Effect,
   Exit,
+  Fiber,
   FiberHandle,
   Layer,
   Option,
   Predicate,
   Queue,
+  Ref,
   ReadonlyArray,
   References,
   Result,
@@ -200,6 +202,8 @@ interface Options {
     }
     readonly hooks?: {
       readonly localPushAdmitted?: (events: ReadonlyArray<LiveStoreEvent.Client.EncodedWithMeta>) => Effect.Effect<void>
+      readonly backendPullCursorAdvanced?: (head: EventSequenceNumber.Client.Composite) => Effect.Effect<void>
+      readonly backendPullRestartRequested?: () => Effect.Effect<void>
     }
   }
 }
@@ -250,6 +254,10 @@ export const make = Effect.fnUntraced(function* ({
   const reservedLocalPushItems = new Set<LocalPushQueueItem>()
   // Ensures mutual exclusion between local push and backend pull processing.
   const localPushBackendPullMutex = yield* Semaphore.make(1)
+  // Pull replacement may interrupt transport work, but never canonical application. This poison-agnostic
+  // boundary also preserves the original application failure when retirement and failure race.
+  const backendPullApplicationFence = yield* Semaphore.make(1)
+  const backendPullRetirementRequested = yield* Ref.make(false)
   // Serializes validation, queue admission, and prefix-fence reconciliation.
   const pushAdmissionSemaphore = yield* Semaphore.make(1)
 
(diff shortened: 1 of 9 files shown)


packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -217,6 +217,9 @@ export const make = Effect.fnUntraced(function* ({
 }: Options) {
   const stateHead = yield* StateHead.StateHead
   const syncBackendPushQueue = yield* TxQueue.unbounded<LiveStoreEvent.Client.EncodedWithMeta>()
+  // A dropping capacity-one signal coalesces repeated requests while the single pull owner
```
slashv: Rephrase this comment in terms easier to understand.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -684,13 +687,15 @@ export const make = Effect.fnUntraced(function* ({
           })
           const error = pushResult.failure
           if (error._tag === 'ServerAheadError') {
-            // It's a core part of the sync protocol that the sync backend will emit a new pull chunk alongside the ServerAheadError
```
slashv, line 687: Why did we delete this comment? is it no longer relevant? If it's a core part, did we change a core part?

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -702,6 +707,10 @@ export const make = Effect.fnUntraced(function* ({
         // This is needed to narrow the Error type. Our retry policy runs indefinitely, but Effect.retry does not narrow the Error type.
         Effect.catchIf((error) => error._tag === 'IsOfflineError' || error._tag === 'UnknownError', Effect.die),
       )
+
+      // The rejected batch remains represented by sync-state pending. Stop this sole drain
```
slashv: Rephrase this comment to make it easier to understand.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -828,9 +836,26 @@ export const make = Effect.fnUntraced(function* ({
         // Needed to avoid `Fiber terminated with an unhandled error` logs which seem to happen because of the `Effect.retry` above.
         // This might be a bug in Effect. Only seems to happen in the browser.
         Effect.provideService(References.UnhandledLogLevel, undefined),
-        Effect.forkScoped,
       )
 
+      // Pull has one owner. A ServerAhead signal retires the old generation before a
```
slashv: The complexity we are adding here makes me second guess this approach. Wondering if there is a more elegant broader solution worth exploring

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
slashv: Rephrased this in terms of the ordering guarantee: stop pushing so later events cannot skip the rejected batch; pull reconciliation rebases pending events and restarts pushing.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
slashv: Agreed. I replaced the second FiberHandle and restart-manager fiber with one supervisor loop that races the active pull against the restart request. Effect interrupts and awaits the losing pull before the loop starts the replacement, preserving the single-pull invariant with substantially less lifecycle machinery.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
slashv: Rephrased this to explain the behavior directly: one pending restart request is enough, and extras can be dropped until it is handled.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
slashv: Good catch. The old comment described the assumption that caused the stall: ServerAhead does not guarantee that the current live pull will publish the missing events. I restored a comment that states the corrected contract and why we start a fresh pull.

slashv:
@codex

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -828,9 +836,25 @@ export const make = Effect.fnUntraced(function* ({
...
-        Effect.forkScoped,
       )
 
+      // A restart request interrupts the current pull before starting another one from
+      // the persisted cursor, so two live pulls can never overlap.
+      const backgroundBackendPullingSupervised = Effect.gen(function* () {
+        while (true) {
+          const outcome = yield* Effect.raceFirst(
+            backendPullingEffect.pipe(Effect.as('pull-ended' as const)),
+            TxQueue.take(backendPullRestartQueue).pipe(Effect.as('restart-requested' as const)),
```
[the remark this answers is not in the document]
slashv: Fixed in 572870b2e. Pull application and retirement now share an explicit application fence plus a retirement flag. A restart marks the generation for retirement, waits for any canonical application already inside the fence to finish, releases the fence before interrupt/cleanup, then awaits and inspects the generation exit before replacement; later applications self-stop. The deterministic regression pauses immediately after Eventlog.updateBackendHead, triggers ServerAhead, proves no replacement starts during that window, and then proves the complete rebase/materialization/state update converges. A second regression proves a terminal canonical application failure wins this race and is not masked by replacement.

packages/@livestore/common/src/leader-thread/LeaderSyncProcessor.ts
```diff
@@ -828,9 +836,25 @@ export const make = Effect.fnUntraced(function* ({
...
+          )
+
+          if (outcome === 'pull-ended') return
```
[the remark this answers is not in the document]
slashv: Fixed in 572870b2e. The single pull owner now outlives a normally completed finite generation: after success it parks on the restart queue, and a later ServerAhead request starts one replacement from the persisted cursor. The deterministic finite-pull regression waits for the initial pull to complete, reproduces accepted-push/lost-publication plus stale-parent ServerAhead, observes the second pull from the persisted root cursor, and verifies the rebased event is pushed.

Issue #1577, opened 2026-08-22
Align sync error taxonomy with supervised processor recovery


slashv:
## Problem

LiveStore's sync error contract and processor behavior disagree about
`UnknownError`:

- `context/02-system/03-sync/spec.md` classifies `UnknownError` as a defect to
  surface rather than retry.
- `context/02-system/03-sync/02-processors/requirements.md` classifies
  offline/unknown failures as transient backend-push errors.
- `LeaderSyncProcessor` retries backend-push `UnknownError` indefinitely, but
  backend pull retries only `IsOfflineError`.
- With the default `onSyncError: 'ignore'`, a non-retried pull or processor
  failure can be logged and then leave that background processor stopped while
  the Store continues with stale data.

Provider transports also map many unrelated expected and unexpected failures
into `UnknownError`. The processor therefore cannot reliably distinguish
connectivity loss, authorization or rate limits, protocol failures, malformed
responses, and internal defects.

This produces two unsafe extremes: a permanent failure can retry forever, while
another failure can silently stop synchronization.

## Expected invariants

- Retry policy is derived from an explicit internal recovery category, not
  from an unclassified catch-all error.
- Only positively identified retryable failures are retried automatically.
- `ServerAheadError` remains a reconciliation signal, not a generic retry.
- Unclassified defects do not enter an infinite retry loop.
- No sync worker terminates silently while the Store continues to appear
  healthy; processor lifecycles are supervised and end in an explicit internal
  recovery or terminal state.
- Applications are not required to implement the sync state machine.

## Suggested scope

1. Reconcile the parent sync error taxonomy with processor requirements through
   an intent decision and a tracked implementation delta.
2. Introduce or refine internal error categories around recovery semantics
   (retryable transport, reconciliation required, backend identity mismatch,
   deterministic/terminal defect).
3. Update providers to preserve positively identified expected failures rather
   than collapsing them into `UnknownError`.
4. Supervise push, pull, and local-apply processors so `onSyncError: 'ignore'`
   cannot silently terminate progress.
5. Add deterministic tests for retryable recovery, terminal defects, and worker
   lifecycle behavior.

## Out of scope

- Public per-commit receipts or application-facing recovery state machines.
- Retry schedule configurability, tracked by #1111.
- `ServerAheadError`/lost-pull liveness recovery, tracked by #1462.
- Crash atomicity and persist-before-admit durability, including #1532.

PR #1580, opened 2026-08-22
Prevent silent sync exits by parking terminal failures

slashv:
## Problem

Backend push retried `UnknownError` forever even though it carries no evidence that replay is safe. Separately, the shared `onSyncError: 'ignore'` handler returned after a terminal cause, so backend push, backend pull, or local apply could lose its background worker while the Store remained open.

This is PR 4 of 4 in native GitHub Stack #1578. It implements the intent-only contract in #1579 on top of #1576's active ServerAhead pull owner.

## Solution

- Retry backend push only for positively classified connectivity failures such as `IsOfflineError`; `UnknownError` reaches terminal supervision after one attempt.
- Preserve #1576's active `ServerAheadError` path: fence the stale push, request persisted-cursor catch-up, retire only between canonical applications, keep one persistent pull owner, and never overlap generations.
- Route generic terminal causes from backend push, backend pull, and local apply through one named supervision policy.
- Under `onSyncError: 'ignore'`, log at error level and park generic terminal work with its unresolved prefix intact. Interrupt-only scope shutdown still completes normally, and a more-specific lifecycle-fatal policy may be inserted ahead of generic parking.
- Keep generic backend-pull parking inside the persistent pull owner, allowing a later ServerAhead request to replace it from authoritative state.
- Retain pull reconciliation's ability to clear, reseed, and replace a parked backend-push worker.
- Preserve the dedicated backend-identity policy and add no public worker-health/status API, callback, receipt, retry configuration, or application recovery state machine.

The combined deterministic regression parks backend pull on `UnknownError`, derives `ServerAheadError` naturally from a stale push, proves exactly one replacement pull starts from the persisted root cursor, proves at most one pull generation is active, and observes processing resume through the rebased push.

Trade-off: generic ignore mode keeps the Store open while terminal local apply remains parked when no independent recovery signal exists. A specialized lifecycle-fatal family may override this generic policy in its owning stack.

## Validation

- `devenv tasks run ts:check`
- `devenv tasks run test:integration:sync-provider:mock`
- `pnpm exec vitest run packages/@livestore/common/src/sync/mock-sync-backend.test.ts tests/package-common/src/leader-thread/LeaderSyncProcessor.test.ts tests/package-common/src/client-session/ClientSessionSyncProcessor.test.ts tests/package-common/src/intent-layer/intent-layer.test.ts` (72 passed, 1 skipped)
- `pnpm exec changeset status`

### Recovery flow

```text
parked backend pull
        ^
        | UnknownError (generic ignore)
stale push -> ServerAhead -> persisted-cursor restart request
                              |
                              v
                 one replacement pull generation
                              |
                              v
                    reconcile + resume push
```

## Related issues

- #1577
- ServerAhead liveness is owned by #1462 and #1575/#1576.
- Retry schedule configurability (#1111), crash atomicity (#1532), public receipts, and worker-health APIs remain out of scope.

### DIFF
--- a/.changeset/quiet-sync-workers.md
+++ b/.changeset/quiet-sync-workers.md
@@ -1,0 +1,8 @@
+---
+'@livestore/common': patch
+---
+
+Retry backend sync operations only after positively identified offline
+failures. `UnknownError` now enters a logged terminal worker state instead of
+being retried indefinitely or letting one sync worker terminate silently under
+`onSyncError: 'ignore'` (#1577).
--- a/.changeset/quiet-sync-workers.md
+++ b/.changeset/quiet-sync-workers.md
@@ -3,6 +3,6 @@
 ---
 
 Retry backend sync operations only after positively identified offline
-failures. `UnknownError` now enters a logged terminal worker state instead of
-being retried indefinitely or letting one sync worker terminate silently under
-`onSyncError: 'ignore'` (#1577).
+failures. `UnknownError` now enters a logged generic terminal worker state
+instead of being retried indefinitely or letting one sync worker terminate
+silently under `onSyncError: 'ignore'` (#1577).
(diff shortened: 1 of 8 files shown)


slashv:
@codex

PR #1598, opened 2026-09-02
fix(common): treat void materializer return as a no-op

xia-chao:
## Summary
- Materializers that return `void`/`undefined` are now a no-op (same as `[]`), instead of reading `.sql` on `undefined` and shutting down the store.
- Docs no longer claim materializers can return an `Effect`. Returning an `Effect` is still unsupported.

## Test plan
- [x] `pnpm exec vitest run src/materializer.test.ts` in `tests/package-common` (8 passed)
- [x] Review changeset `.changeset/fix-void-materializer-noop.md`

### DIFF
--- a/.changeset/fix-void-materializer-noop.md
+++ b/.changeset/fix-void-materializer-noop.md
@@ -1,0 +1,6 @@
+---
+'@livestore/common': patch
+'@livestore/livestore': patch
+---
+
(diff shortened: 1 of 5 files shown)


xia-chao:
Closing: this PR was opened with the wrong GitHub account (`bun-unsafe`). Will resubmit from the intended contributor account.

Issue #1601, opened 2026-09-03
DO<>DO live sync on Cloudflare persistent stubs


bohdanbirdie:
## Problem

DO-RPC live pull keeps a subscription row on the sync backend and calls the client DO back by address
(`bindingName` + DO id). Three gaps follow, and none can be closed on our side:

1. No teardown on ungraceful stop. A DO address never dies, so the backend cannot tell a dead subscriber
   from a hibernating one. Rows leak, dead clients get woken and billed. (#1418 DO-RPC half, #1462)
2. Self-asserted caller identity. The client sends its own DO id, the backend trusts it.
3. A real RPC stub cannot be held: it pins the client DO awake.

We were waiting on cloudflare/workerd#6087 for a hibernation-safe stub.

## Finding

workerd already ships one, undocumented: **persistent stubs**. `ctx.restore(params)` returns a stub whose
target is re-derived by the DO's `[restore](params)` on every call. It can be stored in DO storage as a
token; nothing stays alive while idle. Compat flag `allow_irrevocable_stub_storage` on both Workers.

- landed 2026-05-02: https://github.com/cloudflare/workerd/commit/7d7bfa8da
- experimental gate removed 2026-06-24, still called interim: https://github.com/cloudflare/workerd/commit/a27c16011
- flag and revocation caveat: https://github.com/cloudflare/workerd/blob/a6919c3d0caf/src/workerd/io/compatibility-date.capnp#L1285-L1298
- lazy per-call restore: https://github.com/cloudflare/workerd/blob/a6919c3d0caf/src/workerd/api/worker-rpc.c++#L894-L907
- live in production: https://github.com/cloudflare/workerd/issues/6854
- typed in `@cloudflare/workers-types` 5.x, absent from the 4.x line we pin; no mention in cloudflare-docs or workers-sdk

## Change

Effect RPC stays. Only the live-update leg changes: the client mints a persistent stub for its live pull and
passes it alongside the RPC call, the backend stores it in the subscription row and calls it on publish. The
client's `[restore]` refuses a stale subscription, which drops the row. `bindingName`, `callerContext`,
`syncUpdateRpc` and the request-id matching from #1551 go away.

Two rules, verified on the edge: dispose every stub after use, and signal refusal by return value, never by
throwing in `[restore]` (arrives as a retryable "Network connection lost").

## Evidence

- isolated probe, local + deployed: no pinning, delivery after both sides evicted, stale row dropped, crash recovery
[...]

PR #1602, opened 2026-09-03
Deliver DO-RPC live updates through Cloudflare persistent stubs

bohdanbirdie:
> Draft. Tracks #1601. The Cloudflare API this builds on is live but undocumented; this PR is kept in sync until Cloudflare documents or replaces it.

## Problem

DO-RPC live pull keeps a subscription row on the sync backend and delivers updates by calling the client Durable Object back by binding name and self-reported DO id. A DO address never dies, so the backend cannot tell a dead subscriber from a hibernating one: rows leak and dead clients are woken and billed on every publish (#1418 DO-RPC half, #1462). The id is forgeable. And a real RPC stub could not be held without pinning the client awake.

## Solution

Cloudflare persistent stubs (`ctx.restore` + a `[restore]` method on the DO, compat flag `allow_irrevocable_stub_storage` throughout the restore chain). A stub minted this way can sit in DO storage as a token; nothing stays alive while idle, and the target is re-derived on every call.

Effect RPC is untouched. Only the live-update leg changes:

- Client: a live pull mints `ctx.restore({ storeId, subscriptionId })`, records the id as the store's active subscription in client KV, passes the stub as the second native argument of `rpc(payload, callback)`, and disposes its copy. Graceful shutdown clears the marker and sends `Unsubscribe { subscriptionId }`.
- Backend: the Pull handler stores the row `rpc-sub:<subscriptionId>` with the stub inside. Fan-out delivers each chunk through the stub, disposes it, drops the row on `{ refused: true }`, keeps it on real errors.
- Client DO contract: `[restore]` returns `restoreStoreDoSyncTarget(ctx, params, { onUpdate })`. The target refuses when the id is no longer the store's active subscription, runs `onUpdate(storeId)` so a rebuilt DO can reload its store, refuses again if that reload started a newer live pull (so the row that woke the DO is dropped in the same publish), else routes the bytes into the pull queue as before.
- Removed: `bindingName`, `callerContext`, `ClientDoWithRpcCallback`/`syncUpdateRpc`, and the request-id matched unsubscribe from #1551 (ids are unique per pull now).

Two rules the code follows, both verified on the deployed edge: dispose every stub after use (an undisposed stub keeps both DOs awake), and never signal refusal by throwing inside `[restore]` (it arrives as a retryable "Network connection lost"; the refusal is a return value).

## Cloudflare API status

Rechecked against workerd on 2026-09-12. The persistent-stub contract has not changed since this PR's first edge run, and it remains absent from Cloudflare's public Durable Objects documentation.

- `allow_irrevocable_stub_storage` is explicitly temporary and inherently insecure across trust boundaries: stored capabilities cannot yet be audited or revoked. The storing Worker, target Worker, and every restore-chain member must opt in.
- Turning the flag off breaks existing stored stubs. Migrations must delete backend `rpc-sub:*` rows while the flag remains enabled; client `livestore-rpc-sub:*` markers may be cleared at the same time.
- Configs with compatibility dates before 2026-01-20 also enable `rpc_params_dup_stubs`, matching current RPC ownership semantics so the caller retains the copy it explicitly disposes.
- Stable Workers types still omit `ctx.restore`; the experimental entry exposes it as `any`, and neither entry exports the `restore` symbol. The local checked interface and module augmentation remain necessary.
- workerd#7266 landed retry-neutral Durable Object call machinery, and workerd#7323 is preparing JS-RPC call plans for future retries. No JS-RPC retry behavior has shipped yet. Before adopting a runtime that enables it, this live-update leg needs duplicate-delivery qualification for ambiguous disconnects.

## Trade-offs

- Toolchain: repository Wrangler consumers exercising persistent stubs use 4.128.0 (Miniflare 5 alpha, workerd 1.20260831). The two affected examples now use the same version instead of 4.42.2, whose workerd predates `ctx.restore`.
- `@cloudflare/workers-types` stays on 4.x; `restore` is bridged by a module augmentation in `restore-sync-target.ts`, and `makeDoRpcSync` checks the capability at runtime and dies with a message naming the flag.
- Miniflare 5 locks the shared `.wrangler` state, so sync-provider test files that start a Cloudflare dev server run with `--no-file-parallelism`. `WranglerDevServer` still needs a per-server persist directory before this can be parallelized again.

## Validation

After rebasing onto `main` (`648b8539b`), the implementation passed:

- `devenv tasks run ts:check`
- `vitest run` for common-cf DO-RPC, sync-cf observability, and intent-layer suites: 19/19 pass
- `devenv tasks run test:integration:sync-provider:cf-do-rpc-do`: pass
- `devenv tasks run test:integration:sync-provider:cf-ws-do`: pass

For the 2026-09-12 compatibility correction:

- `devenv tasks run lint:full:fix`: pass
- `devenv tasks run ts:check`: pass
- `devenv shell -- mono test unit`: 87 files, 793 passed, 5 skipped
- Full integration, docs, and example validation: delegated to the fresh PR CI run

The original deployed run used real stores (20 client DOs): both clients and the backend evicted while subscribed; a rebuilt client caught up through `[restore]`; graceful shutdown dropped the row at once; a superseded row was dropped on the next publish; `ctx.abort` recovery and 20/20 fan-out succeeded. Steady delivery was ~60 ms (same as address-based); cold wake of 20 evicted clients with store reload was ~5 s. Worker logs showed every outcome `ok`, a `Dropped refused RPC subscription` line per superseded row, and zero "Network connection lost" errors.

### Demo

Edge lifecycle run, driver summary (two clients A/B, one store, 30 s idle between steps):

```json
{
  "aEvictedWhileSubscribed": true,
  "bEvictedWhileSubscribed": true,
  "backendEvictedWhileSubscribed": true,
  "aRebuiltAndDelivered": true,
  "aEvictedAgain": true,
  "subscriptionRowsAfterT2": 2
}
```

## Remaining before ready

- `WranglerDevServer` per-server persist directory
- Probe: backend DO in a separate Worker; stored stub across a redeploy
- Duplicate-delivery qualification before adopting future automatic JS-RPC retries
- Cloudflare's stance on the API (undocumented, flagged interim)

## Related issues

- #1601 (tracking)
- #1418, #1462, #1415, #1551
- cloudflare/workerd#6087, cloudflare/workerd#4864, cloudflare/workerd#6854
- cloudflare/workerd#7266, cloudflare/workerd#7323


bohdanbirdie:
@bohdanbirdie 
Might ask agent if we can contribute the DO-DO subscription setup we have to the Effect for CF-specific case. One can assume that effect team will be open to that given the raising CF popularity

bohdanbirdie:
Compatibility follow-up pushed in `f505aca1a`.

### Changes since the last iteration

- Rechecked the persistent-stub surface against current workerd. `ctx.restore` / `[restore]` has not changed since the first edge run, but [`allow_irrevocable_stub_storage`](https://github.com/cloudflare/workerd/blob/10c0673595ff30fc520e5d0da1c74be9b68b6977/src/workerd/io/compatibility-date.capnp#L1285-L1304) is still explicitly temporary and inherently insecure across trust boundaries. The storing Worker, target Worker, and every restore-chain member must opt in.
- Added `allow_irrevocable_stub_storage` to both affected examples and the adapter integration fixture, which were already using `[restore]` but had omitted the flag.
- Added `rpc_params_dup_stubs` to the examples and live-pull test fixtures because their compatibility dates predate 2026-01-20. This matches Cloudflare's [current RPC ownership model](https://developers.cloudflare.com/workers/runtime-apis/rpc/lifecycle/) and the code's explicit disposal of the caller's copy.
- Bumped `cloudflare-todomvc` and `web-email-client` from Wrangler 4.42.2 to the repository-tested 4.128.0. The old bundled workerd predates `ctx.restore`; the lockfile now resolves both examples against workerd `1.20260831.1`.
- Corrected the type notes: stable Workers types still omit `DurableObjectState.restore`; the experimental entry types it as `any`, and neither entry exports the `restore` symbol. The checked local interface and module augmentation remain needed.
- Updated the intent specs, decision record, docs, changeset, and changelog with the restore-chain requirement and migration rule: delete backend `rpc-sub:*` rows while the flag is still enabled before turning it off. Existing stored stubs stop working when any restore-chain Worker disables the flag.
- Recorded the new upstream retry risk. [workerd#7266](https://github.com/cloudflare/workerd/pull/7266) landed call-type-neutral DO retry machinery, while [workerd#7323](https://github.com/cloudflare/workerd/pull/7323) is preparing JS-RPC call plans. No automatic JS-RPC retry behavior has shipped yet; duplicate-delivery qualification is now listed before adopting a runtime that enables it.

### Validation for this follow-up

- `devenv tasks run ts:check`
- `devenv shell -- mono test unit` — 87 files, 793 passed, 5 skipped
- Pre-commit `check-quick`

The fresh PR CI run is the source of truth for the full integration, docs, and example matrix. The PR remains draft; the remaining readiness items are unchanged apart from the added duplicate-delivery qualification.

bohdanbirdie:
Rebased onto `main` at `ba85edffd`, which includes #1624 and #1625.

### Changes since the last iteration

- **Spec conflict in `03-cf/spec.md`.** Fan-out now describes both paths: the WebSocket `Exit`/`Defect` cleanup from #1625, and the persistent-stub delivery from this PR. The "WS `Interrupt` emits no Exit" gap is removed, because #1624 fixed it.
- **Lockfile.** Regenerated on top of `main`. Wrangler stays at 4.128.0.

The 9 changed projects type-check clean with `tsgo --build`, and the `common-cf` DO-RPC unit tests pass. I did not rerun the edge lifecycle run.

### Cloudflare status, rechecked 24 Sep

Nothing changed that breaks this PR or unblocks #1601.

- **Unchanged.** The `allowIrrevocableStubStorage` definition in `compatibility-date.capnp` is byte-identical to the 12 Sep recheck. It is still flag-only, with no compat date. The `ctx.restore`/`[restore]` code has only a tracing change (cloudflare/workerd@d735ee3). cloudflare/workerd#6087 has no new activity.
- **Still undocumented.** No docs page, changelog entry, or cloudflare-docs PR mentions the flag or `ctx.restore`. `@cloudflare/workers-types` still types `restore` only in `experimental`, as `restore(params: any): Promise<any>`, so the type bridges stay.
- **JS-RPC retries moved closer.** cloudflare/workerd#7323 merged. Follow-up commits add retry metadata and a pre-dispatch claim check for DO RPC. Everything still sits behind the observe-only `DURABLE_OBJECT_RETRIES_JSRPC` autogate, with no replay loop yet. Arguments that contain an RPC target are classified as never replayable, so the client's `rpc(payload, callback)` call is safe. The backend's delivery call through a restored stub carries only bytes and could become replayable later. The duplicate-delivery qualification in the known gaps stays the open item.
- **New eviction flag to watch.** cloudflare/workerd@d02b8e4 adds an opt-in `durable_object_io_tasks_prevent_eviction` flag. Draft docs cloudflare/cloudflare-docs#33454 say that pending DO RPC calls will keep a DO in memory for up to 15 minutes. This supports the rule to dispose every stub after delivery. It can change idle-eviction timing, so the edge lifecycle run needs a rerun when the flag rolls out.
- **Releases.** Wrangler is now at 4.138.0 (workerd 1.20260921.1). The changelogs from 4.128 to 4.138 have nothing on stubs or restore, so the bump is optional.

Issue #1552, opened 2026-08-11
Schema change to a state table strands existing clients: rebuild is gated on tables being absent, and migrateDb no-ops the change


jwiller:
## Summary

Adding a nullable column to a state table permanently strands every client that already holds a materialized store. The store never finishes booting: the persisted state DB is rejected with `PersistedSqliteError` ("Could not read persisted state db") and `GetRecreateSnapshot` never completes, so the app sits on its loading state forever.

It does not recover from a page reload, a dev-server restart, or reverting the schema change. The only cure I found is `resetPersistence`, which deletes the **eventlog** as well as the app database, so it costs any events the sync backend cannot re-serve.

This contradicts the behaviour the code itself documents in two places:

- `schema-management/migrations.js` header: *"State tables: Recreated and repopulated from eventlog (safe, no data loss)"* and *"User-defined tables are rebuilt from eventlog"*.
- `leader-thread/make-leader-thread-layer.js:21`: `// Either happens on initial boot or if schema changes`, directly above a check that cannot detect a schema change.

## Reproduction

1. Create a store with a state table, e.g. `criteria` with columns `id, boardId, label, detail, order, createdAt, updatedAt, deletedAt`.
2. Commit some events so the tables are materialized, and let it persist (OPFS).
3. Add one nullable column to that table's schema: `probeField: Schema.NullOr(Schema.String)`.
4. Reload.

Expected: the state tables are rebuilt and repopulated from the eventlog.
Actual: the store never boots. Reproduced on a brand-new store with a single nullable column, so it is not accumulated dev churn.

## Root cause

Two defects that compound. Either alone would prevent the rebuild.

**1. The rebuild is gated on the state tables being absent, not stale.**

`make-leader-thread-layer.js:54` calls `recreateDb` only when `dbStateMissing === true`, and line 22 defines that as `!hasStateTables(dbState)`. `hasStateTables` (line 119) only checks that the expected table *names* exist in `sqlite_master`. A `criteria` table that lost a column is still a table called `criteria`, so the check passes, `recreateDb` never runs, and neither `migrateDb` nor `rematerializeFromEventlog` is ever reached. The recorded schema hashes are never consulted at this point.

**2. When it does run, `migrateDb` cannot fix a changed table.**

[...]

Issue #63, opened 2024-04-03
Improved schema hashes


schickling:
More intentional definition and derivation of schema hashes with the following in mind

- hashes shouldn't change when the Effect schema lib is changing
- we need a mechanism that allows for additive non-breaking mutation arg schema changes

PR #1604, opened 2026-09-03
fix(common): replace Effect AST hashes with storage fingerprints

### LINKED ISSUE #1555: Schema hash is derived from effect-internal AST serialization — effect v4 upgrade invalidated all state DBs, forcing full rematerialization on first boot
*Filed as a record/trace per [Discord discussion](https://discord.com/channels/1154415661842452532/1536418308105179236) with @schickling — fine to close right away.*

## What happened (observed in production)

Upgrading a production deployment from an effect-v3-era LiveStore build to current `main` triggered a one-time full rematerialization from the eventlog in every server-side LiveStore client DO (adapter-cloudflare) on its first wake under the new deploy. With 2 client DOs and a 4,466-event eventlog at ~11 `rows_written`/event, the two replays wrote ~101k SQLite rows in a single hour — tripping Cloudflare's account-wide free-tier cap of 100k rows_written/day and blocking all DO writes for ~7.5 h until the daily reset.

Full postmortem:
https://github.com/bohdanbirdie/link-bucket/blob/72d82a0d81b522dd208f147b0c30498845734f91/docs/incidents/2026-08-10-do-rows-written-cap-v4-cutover.md

This is **not** a regression of #1089 — the VFS write path is intact, and the replay cost matches that PR's own measured 12-writes-per-event benchmark.

## Root cause (traced in source at 2e4bcfc)

The schema hash function never changed — its **input** did.

1. `Schema.hash` in `@livestore/utils` is `Hash.string(JSON.stringify(schema.ast, null, 2))` — a hash of effect's *internal Schema AST object serialization*. It is explicitly marked as a temporary workaround for the lack of a schema-hashing API in effect (Effect-TS/effect#2719), and the expression has been unchanged since it was introduced in f0a11fa9 (May 2024):
   https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/utils/src/effect/Schema/index.ts#L107-L119

2. `SqliteAst.hash` deliberately reduces each table AST to stable structural facts via `trimInfoForHasing` (names, column types, nullability, defaults, indexes) — *except* for JSON columns, where it embeds `jsonSchemaHash: Schema.hash(column.schema)`:
   https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/schema/state/sqlite/db-schema/ast/sqlite.ts#L106-L135

3. Effect v4 rewrote SchemaAST (LiveStore's migration commit: https://github.com/livestorejs/livestore/commit/7e843e0118440a485ad590b9c0c6e4967985c4b7), so the same logical schema serializes to different JSON. Minimal demonstration (effect 3.22.1 vs 4.0.0-beta.99):

   ```
   // JSON.stringify(Schema.String.ast, null, 2)
   v3: { "_tag": "StringKeyword", "annotations": { "Symbol(effect/annotation/Title)": "string", ... } }
   v4: { "~effect/Schema": "~effect/Schema", "_tag": "String" }
   ```

   We verified the other moving parts are byte-identical across those versions: `Hash.string` (same DJB2 loop) and `Option`'s JSON form (used for column defaults in the trimmed fingerprint). So the hash change flowed exclusively through `jsonSchemaHash` — with zero change to any schema's meaning. Since client-document tables and most app tables contain JSON columns, effectively every real app is affected on its first boot after upgrading across the v3→v4 boundary.

4. The compound hash over all tables (https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/schema/state/sqlite/mod.ts#L58-L61) is embedded in the state DB **file name** via `getStateDbBaseName`:
   https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/schema/state/sqlite/persistence.ts#L11
   (used e.g. in adapter-cloudflare: https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/adapter-cloudflare/src/make-adapter.ts#L65-L71)

5. On boot the new file is empty, so `dbStateMissing` is true and `recreateDb` runs a full eventlog replay:
   https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/leader-thread/make-leader-thread-layer.ts#L171-L176
   Because the file-name keying acts before any table-level comparison, the per-table migration machinery (`SCHEMA_META_TABLE` diffing in `migrateDb`) never gets a chance to limit the blast radius: any compound-hash drift is a full reset of all tables. The previous `state<hash>@vN.db` file is left orphaned in storage.

6. Replay materializes with one dbState commit per event (`CHUNK_SIZE = 100` is only SELECT pagination), which on the Cloudflare DO VFS costs ~11 billed rows per event:
   https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/rematerialize-from-eventlog.ts#L88

## Suggestions (independent layers, in rough order of cost)

- **Golden-hash regression test:** snapshot the `SqliteAst.hash` / `Schema.hash` output for a fixed reference schema. Any output drift fails CI, turning an accidental store-wide invalidation into a deliberate, changelog-documented one ("this release invalidates state DBs — first boot performs a full rematerialization"). This would have flagged the v4 migration at PR time.
- **Stable hash input (the existing TODO):** derive the JSON-column hash from a LiveStore-owned structural projection of the schema (as `trimInfoForHasing` already does one level up for the table AST) or a stable derived form, instead of `JSON.stringify(schema.ast)` — making the hash immune to effect-internal refactors by construction.
- **Cheaper replay:** wrap replay chunks in a single dbState transaction and/or revive the in-memory-boot-then-flush optimization — both already noted as TODOs (https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/leader-thread/recreate-db.ts#L39-L43, https://github.com/livestorejs/livestore/blob/2e4bcfc68f7ddad5696022a10d515a011f5f785a/packages/@livestore/common/src/rematerialize-from-eventlog.ts#L16). Per #1089's bulk benchmark (6 vs 12 writes/event), batching roughly halves the cost; larger transaction scopes should amortize further.
- **Minor:** clean up orphaned `state<hash>@vN.db` files after a successful rematerialization.

For Cloudflare free-tier users, the pre-upgrade forecast is simple: `eventlog rows × ~11 × (number of server-side client DOs)` vs the 100k/day account-wide rows_written cap.

## References

- Discord thread: https://discord.com/channels/1154415661842452532/1536418308105179236
- Postmortem: https://github.com/bohdanbirdie/link-bucket/blob/72d82a0d81b522dd208f147b0c30498845734f91/docs/incidents/2026-08-10-do-rows-written-cap-v4-cutover.md
- Hash workaround introduced: https://github.com/livestorejs/livestore/commit/f0a11fa93b0646bafba57c69ac3e2fddf40748d5
- SchemaAST v4 migration: https://github.com/livestorejs/livestore/commit/7e843e0118440a485ad590b9c0c6e4967985c4b7
- Effect schema-hashing feature request: https://github.com/Effect-TS/effect/issues/2719
- VFS write-path context: https://github.com/livestorejs/livestore/pull/1089

bohdanbirdie:
## Problem

LiveStore uses the state-schema fingerprint in the persisted database name. The previous fingerprint depended on Effect's internal AST shape and a small numeric hash, so an Effect update could select a fresh database even when storage semantics had not changed. That triggers a full eventlog replay for every store and can add billable Cloudflare writes.

## Solution

- Build an explicit canonical descriptor from LiveStore's table, column, default, and index fields. New LiveStore AST fields must be classified before TypeScript accepts the code.
- Include the public encoded and type representations for declared JSON schemas while ignoring Effect-generated annotation and mutability metadata.
- Hash the canonical bytes synchronously with full-width SHA-256 and encode the result as a 43-character base64url string in every runtime.
- Store textual table/database fingerprints and pin representative outputs with regression tests.
- Keep this implicit: there is no application schema-version, algorithm, override, or migration setting.

## Boundaries

- This removes dependence on Effect's private runtime AST; it does not guarantee that Effect's public schema representation will never evolve. Golden tests make representative drift explicit during upgrades.
- JavaScript callback behavior is not hashed. Custom transformations with the same encoded/type sides, unrepresented `Schema.declare` callbacks, and different runtime-default thunks can remain indistinguishable. The missing Effect-side persistence representation for encoding behavior is tracked in [Effect-TS/effect#8052](https://github.com/Effect-TS/effect/issues/8052).
- Any compound fingerprint change still opens a fresh state database and replays the full eventlog, even if only one table changed.
- This does not choose the separate format-versioning direction explored in #1308.

## Validation

- TypeScript check and full unit suite passed before the rebase.
- Post-rebase fingerprint and intent-layer suites: 21 tests passed.
- Changed TypeScript files: 0 oxlint warnings/errors.
- Three fresh Miniflare/workerd isolates produced the same compound fingerprint as Bun.

Local Bun 1.3.13 benchmark: 30 fresh processes using 30 tables, 450 columns, and 15 distinct JSON schemas. Benchmark sources remain local.

| p50 metric | Before | After |
| --- | ---: | ---: |
| Cold `makeState` latency | 1.80 ms | 5.53 ms |
| Warm `makeState` latency | 0.768 ms | 0.513 ms |
| Retained JSC heap after cold `makeState` and GC | 74 KiB | 396 KiB |

The current `main` generation step blocks the post-rebase full unit task before tests start (`genie:run`: 84 generated files fail). This PR does not modify those generated files.

## Related

- Closes #1555
- Addresses the stable state-schema hashing part of #63; additive event-argument schema evolution remains separate.
- Related to #1552: semantic state-table changes must still rematerialize from the eventlog.
- Related to #1308


bohdanbirdie:
@codex review this PR please

schickling (approved):
Nice work

Image

bohdanbirdie:
Final local stress pass (benchmark sources remain uncommitted).

The application baseline is 30 tables, 450 columns, and 15 distinct typed JSON schemas. Bun results are p50 across 10 fresh processes.

| Metric | Before | After |
| --- | ---: | ---: |
| Cold table hash | 1.125 ms | 1.320 ms |
| Cold `makeState` | 2.082 ms | 6.028 ms |
| Warm table hash | 0.832 ms | 0.539 ms |
| Warm `makeState` | 0.865 ms | 0.562 ms |
| Retained JSC heap after cold state + GC | 941,193 B | 1,269,547 B |
| Process peak RSS | 148,340,736 B | 153,255,936 B |

Five fresh Miniflare/workerd isolates produced deterministic hashes in every sample:

| Workerd metric | Before | After |
| --- | ---: | ---: |
| Cold table hash | 1 ms | 1 ms |
| Cold `makeState` | 3 ms | 5 ms |
| Warm table hash | 1.297 ms | 0.555 ms |
| Warm `makeState` | 1.281 ms | 0.570 ms |
| Benchmark bundle | 382,231 B | 421,351 B |

The local end-to-end Durable Object check used 4,466 events and waited until all 4,466 were durably present before shutdown:

| Boot phase | Todos | Time | Rows written | Rows read | Stored VFS pages |
| --- | ---: | ---: | ---: | ---: | ---: |
| Existing fingerprint reopened | 4,466 | 853 ms | 1 | 84 | 67 |
| First open with the new fingerprint | 4,466 | 1,212 ms | 46,042 | 27,519 | 134 |
| New fingerprint reopened | 4,466 | 61 ms | 0 | 80 | 134 |

The one-time replay wrote about 10.3 tracked SQLite rows per event, consistent with the roughly 11 rows/event reported in #1555; the hashing change does not increase per-event VFS write amplification. The page count doubles because the previous fingerprinted state file remains stored, which is the existing orphan-cleanup concern in #1555. Subsequent same-fingerprint boot wrote zero rows.

A quota-limited deployed check used one temporary personal-account Worker and 1,000 durably synced events. The Worker was deleted after the run.

| Deployed phase | Todos | Client boot | CF CPU | CF wall | Rows written | Rows read | VFS pages |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Legacy fingerprint, settled restart | 1,000 | 303 ms | — | — | 0 | 36 | 23 |
| First open with current fingerprint | 1,000 | 1,878 ms | — | — | 10,110 | 6,061 | 46 |
| Replay CPU probe under another safe 43-character fingerprint | 1,000 | 1,739 ms | 1,615 ms | 1,739 ms | 10,110 | 6,061 | 69 |
| Same probe fingerprint restarted | 1,000 | 149 ms | 115 ms | 150 ms | 0 | 36 | 69 |

No new deployed-runtime issue was found. All 1,000 todos remained correct after replay, changing the fingerprint caused exactly one replay without additional per-event write amplification, and restarting the same fingerprint wrote nothing. The only storage increase was the already-known retained state-file page set tracked in #1555. Cloudflare memory telemetry is aggregated outside live-tail events and was not available from this short-lived isolated deployment, so no deployed memory number is claimed; the local memory stress run found no monotonic retention.

Additional stress coverage:

- 25,000 generated schema fingerprints produced 25,000 distinct values.
- All 14 tested storage-semantic changes changed the fingerprint; table order, index declaration order, and irrelevant Effect metadata did not.
- The resulting Cloudflare filename is 53 characters, below the 56-character limit, and contains only Base64URL-safe characters.
- No monotonic retained heap or object growth appeared over 500 fresh fixture-build-and-hash cycles. The new projection does reach a higher RSS allocator high-water mark, especially for the pathological all-distinct-JSON fixture.
- The known arbitrary-transformation identity limitation remains tracked in Effect-TS/effect#8052.

The interruption test also exposed a separate pre-existing correctness bug: a partially rematerialized state DB can be treated as complete after restart. This is now tracked independently in #1605 and is not being fixed in this PR.

PR #1616, opened 2026-09-07
Fix store.commit callbacks by collecting events before materialization

### LINKED ISSUE #1611: Fix documented store.commit callback failing before materialization
### Problem

The documented callback form of `store.commit` throws before materializing an event:

```ts
store.commit((commit) => {
  commit(events.created({ id: 'todo-1', title: 'item 1' }))
})
```

Observed in the real Cloudflare adapter integration fixture while preparing replay measurements for #1555, on parent commit `6e3e8b1d9633890a83d6520c3c44b6e348fab6aa`:

```text
TypeError: Cannot read properties of undefined (reading 'push')
```

The fixture recorded no attempted materializer calls. Passing events directly (`store.commit(event1, event2)`) succeeds.

### Cause

`packages/@livestore/livestore/src/store/store.ts`, `getCommitArgs`, declares `events` without initializing it, then invokes the callback with `(arg) => events.push(arg)` before assigning the callback's return value to `events`. The documented callback returns void, so initializing the array alone would also need to avoid overwriting it with that return value.

### Expected behavior

Collect callback-emitted events and commit them as the documented transaction. Add regression coverage for a callback emitting one/multiple events, an empty callback and a callback throwing before commit.

This is separate from replay batching. No callback implementation change is included in that work.

bohdanbirdie:
## Problem

Documented `store.commit` callbacks crash before saving any events.

## Solution

Collect callback events before materialization and correct the callback types, including commit options. Throwing callbacks save nothing.

**Breaking:** callback return values are ignored. Replace `store.commit(() => [event])` with `store.commit(event)` or emit through the callback.

## Validation

17 focused tests, the full unit suite (`test:unit`), `ts:check`, and `lint:full:fix` passed.

## Related issues

Closes #1611.

### DIFF
--- a/.changeset/fix-commit-callback.md
+++ b/.changeset/fix-commit-callback.md
@@ -1,0 +1,7 @@
+---
+'@livestore/livestore': minor
+---
+
+Fix the documented `store.commit` callback form and its TypeScript overloads, including calls with commit options. Collect emitted events before materialization so throwing callbacks apply no events. Fixes [#1611](https://github.com/livestorejs/livestore/issues/1611).
+
+Callback return values are now ignored. Replace the undocumented `store.commit(() => [event])` form with `store.commit(event)` or emit events through the supplied callback.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ### Changed
 
+- **Store commits:** Fixed the documented callback form of `store.commit` and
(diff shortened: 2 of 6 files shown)


bohdanbirdie:
@codex

bohdanbirdie:
@bohdanbirdie maybe we can flip this and only support the return of array 
In either way - the less API options for this the better. Need to decide which of the options is better long-term 
Might just flip it

PR #1617, opened 2026-09-07
Fix single-tab independence test by isolating OPFS stores

### LINKED ISSUE #1609: Fix single-tab independence test sharing the same OPFS store
## Problem

The existing `adapter-web > single-tab mode: tabs operate independently` integration test times out waiting for both tabs to boot. Reproduced on unchanged production code at `583bf8420` while validating #1605, as well as with the recovery change. The shared-worker boot test and one-tab fallback test pass.

```sh
devenv shell -- vitest run tests/integration/src/tests/adapter-web/adapter-web.test.ts --testNamePattern 'single-tab mode: tabs operate independently'
```

The assertion at `adapter-web.test.ts` (`expect(boot1 && boot2).toBe(true)`) fails after the 15-second selector timeout.

## Investigation

The fixture disables SharedWorker in both pages and varies their client/session IDs, but `tests/integration/src/tests/playwright/fixtures/adapter-web/Root.tsx` hard-codes the same `storeId` (`adapter-web-test`). Both tabs therefore attempt to open the same OPFS persistence independently. The single-tab adapter explicitly documents that opening multiple tabs with the same store ID can cause data conflicts. This is a likely test setup mismatch; the exact browser failure should be captured before choosing the fix.

Related: #321. Separate from interrupted-rebuild recovery in #1605.

## Checklist

- [ ] Capture the failing page/worker error instead of discarding the selector timeout cause.
- [ ] Decide whether the test should cover separate store IDs or an explicitly supported same-store fallback contract.
- [ ] Adjust the fixture and test accordingly, retaining an assertion that both intended independent stores boot.
- [ ] Run all adapter-web integration tests on the final setup.

Environment: macOS arm64, Node 24, repository Nix/devenv toolchain and Playwright Chromium.

bohdanbirdie:
## Problem

The single-tab independence test opens the same OPFS database twice. Chromium rejects the second handle with `NoModificationAllowedError`, causing a boot timeout.

## Solution

Give each tab a separate store ID and verify that writes remain isolated. Preserve browser errors in boot failures. The shared-worker test keeps its shared store. Test-only change.

## Validation

Reproduced on main. All 3 adapter-web tests, the full unit suite (`test:unit`), `ts:check`, and `lint:full:fix` pass.

## Related issues

Closes #1609.

### DIFF
--- a/tests/integration/src/tests/adapter-web/adapter-web.test.ts
+++ b/tests/integration/src/tests/adapter-web/adapter-web.test.ts
@@ -3,6 +3,7 @@ import { fileURLToPath } from 'node:url'
 
 import { expect } from 'vitest'
 
+import { UnknownError } from '@livestore/common'
 import { BrowserContext, browserContextLayer } from '@livestore/effect-playwright'
 import { CurrentWorkingDirectory, cmd } from '@livestore/utils-dev/node'
 import { Vitest } from '@livestore/utils-dev/node-vitest'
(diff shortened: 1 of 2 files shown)


bohdanbirdie:
@codex

Issue #470, opened 2025-07-14
Embrace Effect Schema for table definitions (schema-first approach)


schickling:
## Overview

Currently, LiveStore uses a **column-first approach** for table definitions where each column is defined separately with its own SQLite type and optionally refined with an Effect schema. This proposal suggests flipping to a **schema-first approach** where the entire table structure is defined using a single Effect schema, with SQLite-specific annotations added as needed.

## Current Implementation (Column-First)

```typescript
const todos = State.SQLite.table({
  name: 'todos',
  columns: {
    id: State.SQLite.text({ primaryKey: true }),
    text: State.SQLite.text({ default: '' }),
    completed: State.SQLite.boolean({ default: false }),
    deletedAt: State.SQLite.integer({ nullable: true, schema: Schema.DateFromNumber }),
  },
})
```

**How it works:**
- Start with SQLite column type (`text`, `integer`, `boolean`, etc.)
- Optionally refine with Effect schema for validation/transformation
- Each column defined independently with its own constraints
- Final row type assembled from individual column definitions

## Proposed Implementation (Schema-First)

```typescript
const todos = State.SQLite.schemaTable({
  name: 'todos',
  schema: Schema.Struct({
    id: Schema.String,
    text: Schema.String,
    completed: Schema.Boolean,
    deletedAt: Schema.optional(Schema.Date),
  }),
  constraints: {
    primaryKey: ['id'],
    defaults: {
      text: '',
      completed: false,
    },
    sqlite: {
      // SQLite-specific annotations
      deletedAt: { columnType: 'integer' }, // Store Date as integer timestamp
    },
  },
})
```

**How it would work:**
- Start with Effect schema defining the complete data structure
- Automatically infer SQLite column types from schema types
- Add SQLite-specific annotations only when needed
- Single source of truth for the data structure

## Type Mapping

The system would automatically map Effect schema types to SQLite columns:

| Effect Schema Type | Default SQLite Column | Notes |
|-------------------|----------------------|-------|
| `Schema.String` | `text` | Direct mapping |
| `Schema.Number` | `real` | Floating point numbers |
| `Schema.Int` | `integer` | Integer numbers |
| `Schema.Boolean` | `integer` | 0/1 representation |
| `Schema.Date` | `text` | ISO string by default |
| `Schema.optional(T)` | `nullable: true` | Nullable columns |
| Complex objects | `text` + JSON | JSON serialization |

## SQLite-Specific Annotations

[...]

Issue #382, opened 2025-06-17
Effect 4 notes


schickling:
TODO when migrating to Effect 4:

- Embrace Effect schema for sqlite state column definitions
  - `State.SQLite.text({ default: '' })` should return `Schema.String` with proper annotations
- Event defs: Use full Effect schema as type param instead of `Encoded` / `Decoded`

IMax153:
Refactor `mono` to use new `ChildProcess` API over the old `Command` module https://github.com/livestorejs/livestore/pull/762#issuecomment-3661827424

PR #1621, opened 2026-09-10
Make Effect Schema the foundation of SQLite table definitions

acusti:
> Spike for the direction discussed in #382: flip the layering so the Effect `Schema.Struct` is the table definition and the SQLite column map is derived from it. The full write-up is in `contributor-docs/rfcs/0004-schema-first-tables.md` on this branch. This is the “parallel PR” that would supercede #1308 that was mentioned in https://github.com/livestorejs/livestore/pull/1308#discussion_r3397345290.

## Problem

Today `SqliteDsl.ColumnDefinition` records are the foundation of a table. `TableDef` is parameterised by that column map, and `table({ schema })` has to translate an Effect schema into it, both at runtime (`getSqlitePropertySignatures`) and at the type level (`SchemaToColumns.FromTypes<Type, Encoded>`, which cannot see codecs or defaults).

That translation layer is where the `v0.5.0-dev.0` upgrade bug in #1597 lives: the type-side AST was substituted for the field's own, stripping a `DateFromString`/`DateFromMillis` codec's encoding, so the column became JSON text that could not be read back. The type level tells the same lie: `rowSchema` reports `createdAt: Schema.Codec<Date, Date>` for a `Schema.Date` field. Beyond that bug, hovers on `typeof tables.users` spell the column map out three times (~4 KB for five columns), `withDefault` on a schema field has no effect on `insert()` typing, and column helpers return records that do not compose with the rest of a user's Effect schemas.

## Solution

A table **is** a `Schema.Struct` whose fields encode to SQLite values. `rowSchema` is that struct; `sqliteDef` (the DDL view used by migrations, `makeColumnSpec` and `astToSql`) and `insertSchema` are derived from it. Nothing downstream of `table()` changed.

- `State.SQLite.text()` & co. return field schemas: `nullable` is `Schema.NullOr`, `default` is an Effect constructor default (so `rowSchema.make` fills it and the type system marks the field `with-default`) plus a DDL annotation; column type, primary key and auto increment are annotations. `table({ name, columns })` is `Schema.Struct(columns)` with a name.
- `TableDef<TName, TFields, TOptions>` replaces `TableDef<SqliteDef, TOptions, TSchema>`. The query builder types read `rowSchema.fields[K]['Type']`. Defaulted and nullable columns are omittable in `insert()` for both column-helper and `withDefault` tables, via `~type.constructor.default` (the v4 feature #382 was waiting on).
- A `Schema.Struct` or `Schema.Class` contributes its `fields` directly, so no property signature is reconstructed and the #1597 class of bug cannot occur on that path. Unions and transformations keep the field's own AST (the #1597 approach, generalised). A bare `Schema.Date` is stored as ISO text, now as the documented normalisation every non-native field goes through (booleans → `0 | 1`, dates → ISO text, else JSON).
- `withDefault` is typed against the field's type, with a `{ sql }` overload. Two docs snippets used `withDefault('CURRENT_TIMESTAMP')` on non-string fields, which stored the literal string; they now use `{ sql: 'CURRENT_TIMESTAMP' }`.

Trade-offs and behaviour changes are listed in the changelog's Breaking Changes entry: helpers no longer return `ColumnDefinition` records, `TableDef`'s generic parameters changed, optional fields are typed `T | null` in the row, and `withColumnType` on a schema field now keeps the field's codec instead of replacing it with the column's default codec.

## Validation

- `vitest run packages/@livestore/common packages/@livestore/livestore`: 291 + package tests pass, including #1597's regression tests and new type-level tests for insert omittability, `rowSchema.make` with value/thunk/SQL defaults, and a canonical-entity-schema table.
- `tsc -b tsconfig.dev.json` (all packages, examples, docs snippets, integration tests): 0 errors.
- The fingerprint tests confirm the derived DDL AST is unchanged for the column-helper DSL.
- `oxfmt` run on the changed files. The repo's oxlint plugin setup was not run (no devenv in the session).
- Not done: `context/02-system/02-state/01-sqlite/spec.md` and the two docs pages still describe the column-map foundation; they should be updated if this lands.

### Demo

The same probe module, declaration-emitted on `main` and on this branch:

```ts
// main (4.2 KB)
export declare const usersBySchema: State.TableDef<State.SqliteTableDefForSchemaInput<"users", {
    readonly id: string; readonly email: string; readonly createdAt: Date; readonly active: boolean;
}, { ...same again... }, Schema.Struct<{ ... }>>, State.TableOptions, Schema.Struct<{
    readonly id: Schema.Codec<string, string, never, never>;
    readonly email: Schema.Codec<string, string, never, never>;
    readonly createdAt: Schema.Codec<Date, Date, never, never>;   // <- not what SQLite stores
    readonly active: Schema.Codec<boolean, boolean, never, never>;
}>>;

// this branch (1.0 KB)
export declare const usersBySchema: State.TableDef<"users", {
    readonly id: Schema.String;
    readonly email: Schema.String;
    readonly createdAt: Schema.Date;
    readonly active: Schema.withConstructorDefault<Schema.Boolean>;
}, State.WithDefaults>;
```

## Related issues

- #382
- #1597 (supersedes the `getSqlitePropertySignatures` half of that fix; keeps the bare `Schema.Date` rule)
- #470
- #1307
- #1308 (PR that resolves 1307)

https://claude.ai/code/session_01D2pDUTNUKWmqen8KFNNUXf

---
_Generated by [Claude Code](https://claude.ai/code/session_01D2pDUTNUKWmqen8KFNNUXf)_


acusti:
i also asked fable to look into the interaction between this and the new storage fingerprinting introduced in #1604:

#1604 made the state-database fingerprint independent of Effect’s AST internals, and this PR changes what a column’s codec is: for fields stored as-is it is now the user’s field schema itself, which carries the column annotations from `text({ primaryKey: true })`, the `withConstructorDefault` context from a `default`, and the optional-key context. If any of that reached the fingerprint, every existing database would rematerialize once on upgrade.

Happily, it doesn’t. The descriptor is built from LiveStore’ own column fields (name, type, primary key, nullable, auto-increment, default); a column’s schema only enters the hash for JSON columns, via `SchemaRepresentation.toRepresentations` with `annotations` stripped. `fingerprint.ts` is untouched by this PR.

I verified it empirically rather than by reading: the same 18 table shapes fingerprinted on `main` and on this branch.

| shape | result |
|---|---|
| column-helper physical table (`text`/`integer`/`boolean`/`datetime`/`real`/`blob`, with defaults and nullables) | identical |
| `json()` bare, `nullable`, typed, typed + nullable, value default, `null` default, thunk default | identical |
| `text({ schema: Schema.fromJsonString(...) })` | identical |
| schema path: struct, `NullOr(struct)`, `optional(struct)`, `struct.pipe(withDefault(...))`, arrays, literal unions, `withDefault` on text/int | identical |
| system tables (`makeState({ tables: [] }).sqlite.hash`) | identical |
| schema path: `DateFromString` / `NullOr(DateFromMillis)` field | **changed** |

The one change is #1597’s bug itself: on `main` that column was JSON text that could not be read back, so the fingerprint changing there is the fix taking effect, and a fresh database is what those users need anyway.

Since #1604 pins only two shapes, and I tripped the system-tables pin myself while iterating (switching `json()`’s runtime schema from `Schema.Any` to `Schema.Unknown` changes the JSON descriptor, which is why the type is `unknown` but the runtime schema stays `Any`), commit 15101916 turns that catalog into a pinned test in `fingerprint.test.ts`. A future change to the definition layer that alters storage identity now fails a test instead of silently rebuilding everyone’s database.

PR #1623, opened 2026-09-12
Simplify store.commit callbacks with array-returned batches

### LINKED ISSUE #1611: Fix documented store.commit callback failing before materialization

bohdanbirdie:
## Problem

[PR #1616](https://github.com/livestorejs/livestore/pull/1616) fixed the documented `store.commit` emitter callback by collecting its events before materialization and explicitly ignoring callback return values. The [post-merge design comment](https://github.com/livestorejs/livestore/pull/1616#issuecomment-5622042810) asked whether callbacks should instead return arrays so LiveStore supports only one callback convention for describing a batch.

The emitter shape does not match the implementation particularly well: events are not committed while the callback runs, and queries in the callback see the pre-commit state. Its `void` return type also cannot enforce the synchronous boundary because TypeScript deliberately permits non-void and async functions where a void-returning callback is expected. That leaves room for silently ignored arrays and for emissions after an `await` to be lost.

## Solution

Replace emitter callbacks with synchronous batch builders that must return a `ReadonlyArray` of schema events:

```ts
store.commit(() => [event1, event2])
```

This keeps one callback convention and models the real operation as "build the complete batch, then commit it." Array returns compose with ordinary array helpers, reject typical async callbacks at the type level, and preserve the existing all-or-nothing behavior when the builder throws. Direct `store.commit(...events)`, commit options, ordering, and empty batches remain supported.

Runtime callers using JavaScript or `any` receive a focused error for non-array results, including promises. LiveStore copies a valid returned array before it enters the commit pipeline. The store intent layer records the choice and rejected alternatives, while docs, tests, the changeset, and the unreleased changelog now describe the final contract.

The trade-off is that imperative emitter callbacks must migrate. Callers that prefer imperative construction can build a local array and return it. Removing callbacks entirely was considered, but retaining a scoped, lazy builder remains useful for conditional batches. A future API exposing reads and writes inside a real transaction should be separate from this batch builder.

## Validation

- `pnpm exec vitest run packages/@livestore/livestore/src/store/store-commit.test.ts` (10 tests passed)
- `devenv tasks run ts:check`
- `pnpm exec tsc -p packages/@livestore/livestore/tsconfig.json --noEmit`
- `devenv tasks run lint:check:md-imports`
- Focused Oxlint for the modified TypeScript files (0 warnings, 0 errors)
- `devenv tasks run release:changeset:check-pr`
- `devenv tasks run release:changeset:check-bodies`
- Full docs package build and internal-link validation
- Pre-commit `check-quick`

`devenv tasks run lint:full:fix` was also run. Its formatter completed, but the aggregate task remains blocked by six pre-existing Oxlint errors in unrelated example files; the modified TypeScript files pass focused Oxlint.

### Demo

```diff
-store.commit((commit) => {
-  commit(event1)
-  commit(event2)
-})
+store.commit(() => [event1, event2])
```

## Related issues

- Original callback failure: [#1611](https://github.com/livestorejs/livestore/issues/1611)
- Previous fix: [#1616](https://github.com/livestorejs/livestore/pull/1616)
- Follow-up API discussion: [#1616 comment](https://github.com/livestorejs/livestore/pull/1616#issuecomment-5622042810)

---

> [!NOTE]
> **Medium Risk**
> This is a breaking change to a core `store.commit` API surface; apps or libraries still on the emitter callback must migrate, though direct event commits and all-or-nothing semantics are unchanged.
> 
> **Overview**
> **`store.commit` callback batches now return event arrays** instead of using an emitter function. Callers should use `store.commit(() => [event1, event2])` (or keep direct `store.commit(...events)`); migrate away from `store.commit((commit) => { commit(event) })`.
> 
> Runtime behavior matches the new shape: the builder runs first, LiveStore **copies** the returned array, then materializes only after a successful return. Throwing still applies nothing; **`[]` is a no-op**. Non-array results (including promises) throw a focused `TypeError`. TypeScript overloads use **`CommitBatchBuilder`** so async callbacks and emitter parameters are rejected at compile time.
> 
> Docs, store spec, ADR **0002**, changeset/changelog breaking notes, and **`store-commit.test.ts`** are updated to describe and lock in the contract (follow-up to [#1616](https://github.com/livestorejs/livestore/pull/1616), fixes [#1611](https://github.com/livestorejs/livestore/issues/1611)).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5d2e0bd1bb9472abcc2653ad120c0da64d3d105. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-commit-callback.md
+++ b/.changeset/fix-commit-callback.md
@@ -2,6 +2,6 @@
 '@livestore/livestore': minor
 ---
 
-Fix the documented `store.commit` callback form and its TypeScript overloads, including calls with commit options. Collect emitted events before materialization so throwing callbacks apply no events. Fixes [#1611](https://github.com/livestorejs/livestore/issues/1611).
+Fix the `store.commit` callback form and its TypeScript overloads, including calls with commit options. A synchronous callback now returns an array describing the complete event batch, which LiveStore collects before materialization so throwing callbacks apply no events. Fixes [#1611](https://github.com/livestorejs/livestore/issues/1611) and follows up on [#1616](https://github.com/livestorejs/livestore/pull/1616).
 
-Callback return values are now ignored. Replace the undocumented `store.commit(() => [event])` form with `store.commit(event)` or emit events through the supplied callback.
+Replace `store.commit((commit) => { commit(event) })` with `store.commit(() => [event])` or `store.commit(event)`. Callback builders must be synchronous and return an array, including `[]` for an empty batch.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,11 @@
 
 ### Changed
 
-- **Store commits:** Fixed the documented callback form of `store.commit` and
-  its TypeScript overloads, including calls with commit options. Callback events
-  are collected before materialization, and throwing callbacks apply no events
-  ([#1611](https://github.com/livestorejs/livestore/issues/1611)).
+- **Store commits:** Fixed the callback form of `store.commit` and its TypeScript
+  overloads, including calls with commit options. Synchronous callbacks return
+  an event array that is collected before materialization, and throwing callbacks
+  apply no events ([#1611](https://github.com/livestorejs/livestore/issues/1611),
+  [#1616](https://github.com/livestorejs/livestore/pull/1616)).
 - **Cloudflare sync telemetry:** Added optional Effect tracer layers via `otel: layer`.
   Telemetry cleanup runs in the background, and finite WebSocket operations and
   DO-RPC pull streams now emit their sync spans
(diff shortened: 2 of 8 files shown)


bohdanbirdie:
@codex

PR #1625, opened 2026-09-12
Clean up terminal WS pull fan-out state

### LINKED ISSUE #1418: Server DO does not tear down live pull streams on client Interrupt (no Exit emit; DO-RPC path is TODO)

bohdanbirdie:
## Problem

The sync Durable Object persists live pull request IDs in each WebSocket attachment for manual push fan-out. It previously removed IDs only when an \`Interrupt\` arrived.

Consequently, finite completion and typed failure left terminal request IDs in the attachment. A connection-level RPC \`Defect\` also terminates every request on Effect's client while leaving all persisted fan-out IDs behind. Later pushes could continue sending \`Chunk\` frames to requests the client had already ended.

## Solution

- Add a narrow per-request \`onRequestExit\` lifecycle hook to the common Cloudflare WebSocket RPC adapter.
- Add a narrow connection-level \`onProtocolDefect\` hook for the global terminal case.
- Route ordinary and reconstructed exits through one exit writer.
- Route Effect, parsing, and encoding defects through one defect writer.
- Remove one persisted pull ID after \`Exit\`; clear all IDs after \`Defect\`.
- Run lifecycle cleanup for buffered serializer output and failed socket writes.
- Catch and log hook failures so callbacks cannot change the protocol result.

Both cleanup paths share one idempotent WebSocket attachment updater.

## API impact

\`DurableObjectWebSocketRpcConfig\` gains two additive optional callbacks:

- \`onRequestExit(requestId, ws)\`
- \`onProtocolDefect(ws)\`

Existing users require no changes. The hooks expose only terminal lifecycle events, and a protocol defect does not force a socket reconnect.

## Validation

- \`devenv tasks run lint:full:fix\`
- \`devenv tasks run ts:check\`
- \`devenv tasks run test:unit\`
- \`devenv tasks run test:integration:sync-provider:cf-ws-do\`
- Targeted hibernated-defect regression: failed with \`['restored-pull']\` before the fix and passed with \`[]\` after.
- Intent-layer invariant suite.
- Independent adversarial subagent review found no remaining blocker.

The regressions establish that restored IDs exist before testing removal, cover normal completion and typed failure, and force a protocol defect after actual Durable Object reconstruction.

## Stack

Depends on #1624, which contains only the hibernated \`Interrupt → Exit\` correction and closes #1418.

## Related issues

Follow-up to #1624 and #1418. No separate existing issue matched this lifecycle cleanup.

### DIFF
--- a/.changeset/calm-websockets-exit.md
+++ b/.changeset/calm-websockets-exit.md
@@ -3,4 +3,4 @@
 '@livestore/sync-cf': patch
 ---
 
-Complete an interrupted WebSocket live pull with its terminal RPC `Exit` after the sync Durable Object hibernates and reconstructs. The Cloudflare WebSocket protocol now preserves Effect interruption semantics when its per-request schema state was lost to hibernation (#1418).
+Complete WebSocket pull lifecycles across sync Durable Object hibernation. Interrupted live pulls now receive their terminal RPC `Exit` after reconstruction, and terminal pulls are removed from persisted fan-out state (#1418).
--- a/.changeset/calm-websockets-exit.md
+++ b/.changeset/calm-websockets-exit.md
@@ -3,4 +3,4 @@
 '@livestore/sync-cf': patch
 ---
 
-Complete WebSocket pull lifecycles across sync Durable Object hibernation. Interrupted live pulls now receive their terminal RPC `Exit` after reconstruction, and terminal pulls are removed from persisted fan-out state (#1418).
+Complete WebSocket pull lifecycles across sync Durable Object hibernation. Interrupted live pulls now receive their terminal RPC `Exit` after reconstruction, and terminal or defected pulls are removed from persisted fan-out state (#1418).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,9 @@
   ([#1537](https://github.com/livestorejs/livestore/pull/1537)).
 - **Cloudflare WebSocket sync:** Live pulls interrupted after the sync Durable
   Object hibernates now receive their terminal RPC `Exit`, matching interruption
-  behavior before hibernation. Completed pulls are also removed from persisted
-  fan-out state ([#1418](https://github.com/livestorejs/livestore/issues/1418)).
+  behavior before hibernation. Completed and protocol-defected pulls are also
+  removed from persisted fan-out state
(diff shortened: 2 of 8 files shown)


bohdanbirdie:
@codex

